Issue #7, opened 2018-01-15
Order of commands in `help`


eordano:
In hierarchy of importance:

* init
* start
* upload
* link
* help

Take out: exit and update-linker

NicoSantangelo:
To achieve this, we need to remove the default help method from [vorpal](https://github.com/dthree/vorpal) and re-write it.

I'm trying to use [vorpal.help()](https://github.com/dthree/vorpal/blob/604a96682e2345da4dda7ebf26c477350a1b9c6d/lib/vorpal-commons.js#L20) as a template to avoid hardcoding the help text (because it's going to be a pain to maintain)

PR #22, opened 2018-01-16
New help command order #7

NicoSantangelo:
This is way harder than it should be.

Duplicating the logic on the existing `help` it's not ideal, but [vorpal](https://github.com/dthree/vorpal) doesn't really help. I tried to leave a compelling comment on the code

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -408,6 +408,59 @@ cli
     linker.bind(cli)(args, this, callback);
   });
 
+ /**
+ * `help` command for a description of all commands or a single one
+ */
+const help = cli.find('help');
+if (help) {
+  help.remove();
+}
+
+cli
+  .command('help [command...]')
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -408,6 +408,59 @@ cli
...
+
+      this.log(cli._commandHelp());
+
+      cli.commands = oldCommands;
+    }
+
+    callback();
+  });
+
+function showHelp(commandName: string) {
```
NicoSantangelo: Reviewer:

Do we have a place to put this? I only see an `utils` folder but adding a cli-related standalone method seems out of place

michaltakac: We will have in ~30 minutes ;)

michaltakac: I'll merge and then move in my next PR.

PR #23, opened 2018-01-16
Chore/commands split

### LINKED ISSUE #19: Split commands from index.ts to their respective files

michaltakac:
Fix #19

### DIFF
--- a/src/commands/help.ts
+++ b/src/commands/help.ts
@@ -1,35 +1,12 @@
-
-
 export function help(vorpal: any) {
-  function showHelp(commandName: string) {
-    commandName = commandName.toLowerCase().trim();
-
-    const command = vorpal.commands.find(
-      (command: any) => command._name === commandName
-    );
-
-    if (command && !command._hidden) {
-      if (typeof command._help === 'function') {
-        command._help(commandName, (str: string) =>
-          vorpal.log(str)
-        );
-      } else {
-        vorpal.log(command.helpInformation());
-      }
-    }
-  }
-
-  const help = vorpal.find('help');
-  if (help) {
-    help.remove();
-  }
+  removeDefaultHelp(vorpal);
 
   vorpal
     .command('help [command...]')
     .description('Provides help for a given command.')
     .action(function(args: any, callback: () => void) {
       if (args.command) {
-        showHelp(args.command.join(' '));
+        showHelp(vorpal, args.command.join(' '));
       } else {
         // vorpal doesn't have a nice way to customize the full help list.
         // It uses a somewhat internal `_commandHelp` ( https://github.com/dthree/vorpal/blob/master/lib/vorpal.js#L1079 ) which is huge and not abstracted.
@@ -50,4 +27,29 @@ export function help(vorpal: any) {
 
       callback();
     });
-};
+}
+
+function removeDefaultHelp(vorpal: any) {
+  const help = vorpal.find('help');
+  if (help) {
+    help.remove();
+  }
+}
+
+function showHelp(vorpal: any, commandName: string) {
+  commandName = commandName.toLowerCase().trim();
+
+  const command = vorpal.commands.find(
+    (command: any) => command._name === commandName
+  );
+
+  if (command && !command._hidden) {
+    if (typeof command._help === 'function') {
+      command._help(commandName, (str: string) =>
+        vorpal.log(str)
+      );
+    } else {
+      vorpal.log(command.helpInformation());
+    }
+  }
+}
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -1,8 +1,8 @@
 import chalk from 'chalk';
 import fs = require('fs-extra');
+import inquirer = require('inquirer');
 import { cliPath }from '../utils/cli-path';
 import { prompt } from '../utils/prompt';
-import { sceneMeta } from '../utils/scene-meta';
 import { generateHtml } from '../utils/generate-html';
 import { isDev } from '../utils/is-dev';
 import { wrapAsync } from '../utils/wrap-async';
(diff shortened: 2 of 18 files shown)


src/commands/init.ts
```diff
@@ -0,0 +1,183 @@
...
+export function init(vorpal: any) {
+  vorpal
+    .command('init')
+    .description('Generates new Decentraland scene.')
+    .option(
+    '-p, --path <path>',
+    'Output path (default is the current working directory).'
+    )
+    .option('--boilerplate', 'Include sample scene.')
+    .action(async function (args: any, callback: () => void) {
```
menduz: action does not work with async funcitons.

consider writing a helper that handles the promise's fullfilment/rejection

```js
function wrap(fn: async (args) => void): (this: any, args: any, callback: () => void) => void {
  return function(args, cb) {
    fn.call(this, args)
      .then(() => cb())
      .catch((err) => {
        // do something with the error
        cb(err);
      });
  }
}





/// Use it like 
vorpal
  .command('init')
  .action(wrap(async function(args) {
    this.prompt(...)
    await something();
  }))
```

src/commands/init.ts
```diff
@@ -0,0 +1,183 @@
...
+        ? tags
+          .split(',')
+          .map(tag => tag.replace(/\s/g, ''))
+          .filter(tag => tag.length > 0)
+        : [];
+
+      self.log(chalk.blue('Contact information:'));
+
+      sceneMeta.owner = await prompt(self, chalk.blue(' your MetaMask address: '));
+      sceneMeta.contact.name = await prompt(self, chalk.blue(' your name: '));
```
menduz: Please make more use of bundled prompt features instead of having lots of `await prompt(...)`

https://www.npmjs.com/package/inquirer

NicoSantangelo:
Got it to work! but I had a little problem along the way. Checking out the branch and running `npm start` didn't work out of the gate

```bash
➜  decentralandia/cli git:(master) git checkout chore/commands-split
Branch chore/commands-split set up to track remote branch chore/commands-split from origin.
Switched to a new branch 'chore/commands-split'

➜  decentralandia/cli git:(chore/commands-split) npm start
> dcl-cli@0.1.4 start /Users/nicosantangelo/Work/decentralandia/cli
> ./dev.js

Decentraland CLI v0.1.4

Type "exit" to quit, "help" for a list of commands.

dcl $
You're now in development mode.

/Users/nicosantangelo/Work/decentralandia/cli/dev.js:7
cli.init();
    ^

TypeError: cli.init is not a function
    at Object.<anonymous> (/Users/nicosantangelo/Work/decentralandia/cli/dev.js:7:5)
    at Module._compile (module.js:573:30)

➜  decentralandia/cli git:(chore/commands-split) npm i
> dcl-cli@0.1.4 prepublish /Users/nicosantangelo/Work/decentralandia/cli
> npm run build

> dcl-cli@0.1.4 build /Users/nicosantangelo/Work/decentralandia/cli
> npm run clean && tsc && npm run linker:build

> dcl-cli@0.1.4 clean /Users/nicosantangelo/Work/decentralandia/cli
> rimraf dist && rimraf tmp

> dcl-cli@0.1.4 linker:build /Users/nicosantangelo/Work/decentralandia/cli
> next build && next export -o dist/linker-app

{ ValidationError: UglifyJs Plugin Invalid Options

options['compress'] should NOT have additional properties

  name: 'ValidationError',
  message: 'UglifyJs Plugin Invalid Options\n\noptions[\'compress\'] should NOT have additional properties\n',
  errors:
   [ { keyword: 'additionalProperties',
       dataPath: '[\'compress\']',
       schemaPath: '#/additionalProperties',
       params: [Object],
       message: 'should NOT have additional properties' } ] }

➜  decentralandia/cli git:(chore/commands-split) npm start

> dcl-cli@0.1.4 start /Users/nicosantangelo/Work/decentralandia/cli
> ./dev.js

You're now in development mode.

Decentraland CLI v0.1.4

Type "exit" to quit, "help" for a list of commands.

dcl $
```

I'm not sure if this is a problem on my end or if we should address it before merging

src/commands/help.ts
```diff
@@ -0,0 +1,53 @@
+
+
+export function help(vorpal: any) {
+  function showHelp(commandName: string) {
```
NicoSantangelo: We're recreating this function each time help is called. We can leave it below the main `help` function

src/commands/help.ts
```diff
@@ -0,0 +1,53 @@
...
+        command._help(commandName, (str: string) =>
+          vorpal.log(str)
+        );
+      } else {
+        vorpal.log(command.helpInformation());
+      }
+    }
+  }
+
+  const help = vorpal.find('help');
```
NicoSantangelo: this could also be extracted so it's a bit less confusing to something like:

```javascript
  removeDefaultHelp(vorpal)
  // (...)
}

function removeDefaultHelp(vorpal) {
  const help = vorpal.find('help');
  if (help) {
    help.remove();
  }
}
```

src/commands/init.ts
```diff
@@ -0,0 +1,183 @@
...
+  vorpal
+    .command('init')
+    .description('Generates new Decentraland scene.')
+    .option(
+    '-p, --path <path>',
+    'Output path (default is the current working directory).'
+    )
+    .option('--boilerplate', 'Include sample scene.')
+    .action(async function (args: any, callback: () => void) {
+      const self = this;
```
NicoSantangelo: We can tackle this later but using this is no longer necessary as we can just use a lambda when needed.
Same for the rest of `const self = this;` along the way

src/commands/upload.ts
```diff
@@ -0,0 +1,116 @@
...
+        content: new Buffer(fs.readFileSync(`${root}/scene.html`))
+      },
+      {
+        path: `tmp/scene.json`,
+        content: new Buffer(fs.readFileSync(`${root}/scene.json`))
+      }
+    ];
+
+    // Go through project folders and add files if available
+    await fs.readdir(`${root}/audio`).then(files =>
```
NicoSantangelo: Again this can be done in a later issue, but we can refactor every `await fs.readdir((...))` as they're all similar and seem to differ only the path.

Also, adding a `await` => `then` is not wrong but it's a bit misleading to me. I would take advantage of the `await` to assign to a new variable and use it below the promise

src/commands/link.ts
```diff
@@ -0,0 +1,12 @@
+import { linker } from '../utils/linker';
+
+export function link(vorpal: any) {
+  vorpal
+    .command('link')
+    .description('Link scene to Ethereum.')
+    .action(function (args: any, callback: () => void) {
+      const self = this;
+
+      linker.bind(vorpal)(args, this, callback);
```
NicoSantangelo: Why are we binding `vorpal` here? it's not used on `serve.ts` and the `vorpalInstance` is already supplied as an argument

src/commands/start.ts
```diff
@@ -0,0 +1,11 @@
+import { serve } from '../utils/serve';
+
+export function start(vorpal: any) {
+  vorpal
+    .command('start')
+    .alias('serve')
+    .description('Starts local development server.')
+    .action(function (args: string, callback: () => void) {
+      serve.bind(vorpal)(args, this, callback);
```
NicoSantangelo: Why are we binding `vorpal` here? it's not used on `serve.ts` and the `vorpalInstance` is already supplied as an argument

src/utils/linker.ts
```diff
@@ -3,10 +3,10 @@ import fs = require('fs-extra');
 import Koa = require('koa');
 import Router = require('koa-router');
 import serve = require('koa-static');
-import isDev from './is-dev';
+import { isDev } from './is-dev';
 import { prompt } from './prompt';
 
-export default async (args: any, vorpal: any, callback: () => void) => {
+export async function linker(args: any, vorpal: any, callback: () => void) {
```
NicoSantangelo: Every other `vorpal` argument is called `vorpalInstance`. We might want to keep things a bit more consistent and rename here (or in every other place)

menduz:
LGTM, squash and merge if it works!

PR #24, opened 2018-01-17
Feature/docker names

### LINKED ISSUE #5: Default project title: Let's play with some docker-style names for the scene

NicoSantangelo:
Fix #5

~~Blocked by #23 !~~

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,6 +8,9 @@
 /// <reference path="../typings/vorpal.d.ts" />
 /// <reference path="../typings/dcl.d.ts" />
 
+// Use custom docker names type definitions until there's official one
+/// <reference path="../typings/docker-names.d.ts" />
+
 import chalk from 'chalk';
 import fs = require('fs-extra');
 import path = require('path');
--- a/src/utils/project.ts
+++ b/src/utils/project.ts
@@ -1,0 +1,9 @@
+import dockerNames = require('docker-names');
+
+export function getDefaultName(): string  {
+  return 'dcl-app';
+}
+
+export function getRandomName(): string {
+  return dockerNames.getRandomName();
+}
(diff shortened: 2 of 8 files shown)


src/index.ts
```diff
@@ -8,6 +8,9 @@
 /// <reference path="../typings/vorpal.d.ts" />
 /// <reference path="../typings/dcl.d.ts" />
 
+// Use custom docker names type definitions until there's official one
+/// <reference path="../typings/docker-names.d.ts" />
```
NicoSantangelo: I left this here instead of on `project.ts` (where it's used) to see every type reference at a glance. Let me know if we want to move it elsewhere

src/utils/project.ts
```diff
@@ -1,11 +1,9 @@
 import dockerNames = require('docker-names');
```
NicoSantangelo: What do you think of this file?

My first approach was to have a `project` object, adding every method as a property to it. I ended up going with different functions to keep the way `export/import` work on the repo.

This way we can do:

```javascript
import { getDefaultName, (...) } from '../utils/project';
// and
import * as project from '../utils/project';
```

if I export a single object we'll need to do:

```javascript
import { project } from '../utils/project';

// elsewhere
const { getDefaultName, (...) } = project
```

Let me know what you think is best

NicoSantangelo: What do you think of this file?

My first approach was to have a `project` object, adding every method as a property to it. I ended up going with different functions to keep the way `export/import` work on the repo.

This way we can do:

```javascript
import { getDefaultName, (...) } from '../utils/project';
// and
import * as project from '../utils/project';
```

if I export a single object we'll need to do:

```javascript
import { project } from '../utils/project';

// elsewhere
const { getDefaultName, (...) } = project
```

Let me know what you think is best

NicoSantangelo:
Rebased!

michaltakac:
Great, thanks!

PR #25, opened 2018-01-17
Feature/help as default command

### LINKED ISSUE #4: `dcl` should not open the interactive console
The interactive console is probably not going to be very useful until some point in the future. Let's hide that and make "help" the default command.

NicoSantangelo:
Fix #4

~~Blocked by #23 !~~

See discussion on https://github.com/decentraland/cli/issues/4 before merging

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -51,15 +51,18 @@ const cli = {
 
     vorpal
       .delimiter(DELIMITER)
-      .show();
+      .catch('[words...]', 'Catches incorrect commands')
+      .action(() => {
+        vorpal.execSync('help');
+      });
 
-    // If one or more command, execute and close
     if (process.argv.length > 2) {
+      // If one or more command, execute and close
       vorpal.parse(process.argv);
     } else {
-      // Enters immersive mode if no commands supplied
+      // Show help if no commands are supplied
       vorpal.log(`Decentraland CLI v${VERSION}\n`);
-      vorpal.log('Type "exit" to quit, "help" for a list of commands.\n');
+      vorpal.exec('help');
     }
   }
 };
--- a/dev.js
+++ b/dev.js
@@ -3,5 +3,5 @@
 'use strict';
 
 var cli = require(".");
-console.log('\nYou\'re now in development mode.\n');
+console.log('\nYou\'re now in development mode.');
 cli.init();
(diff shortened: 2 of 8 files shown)


src/index.ts
```diff
@@ -37,439 +34,36 @@ export const DELIMITER = 'dcl $';
...
-        cli.log(str)
-      );
+    if (process.argv.length > 2) {
+      // If one or more command, execute and close
+      vorpal.parse(process.argv);
     } else {
-      cli.log(command.helpInformation());
+      // Show immersive mode if no commands are supplied
+      vorpal.log(`Decentraland CLI v${VERSION}\n`);
+      vorpal.exec('help');
```
michaltakac: We can disable immersive mode by doing:
`vorpal.exec('help').then(() => vorpal.exec('exit'));`.

NicoSantangelo:
Rebased!

src/index.ts
NicoSantangelo: I think the current version works fine because I removed the call to `.show()`. Running the cli with no arguments shows:

image

michaltakac: I think that when you compile the code with `tsc` and test the `dcl` command, it will work, but `dcl upload` won't work. At least that's what happened for me when I tried removing the `.show()`

NicoSantangelo: Hm can you reproduce it on this branch?
I ask because adding a `vorpal.exec('help').then(() => vorpal.exec('exit'));` shouldn't help either, as `vorpal.exec('help');` only runs if you run the cli without arguments.

Running `upload` gives me:

```bash
➜  decentralandia/cli git:(feature/help-as-default-command) ./dev.js upload                                                                                                                        16:10:56

You're now in development mode.
vorpal.getSessionById was called with no ID passed.
```

:/

michaltakac: Without `.show()`, `init` command hangs:
screen shot 2018-01-17 at 20 58 39

WIth `.show()`, `init` works as expected:
screen shot 2018-01-17 at 20 59 12

NicoSantangelo: Good lord I don't like vorpal. I think I fixed it but i'll let you know on the main chat of this issue

NicoSantangelo:
@michaltakac It should be fixed now, I moved from vorpal to inquirer to avoid the early exit.

Let me know if you find anything else!

michaltakac:
Works now!

Issue #14, opened 2018-01-15
Serve "scene.html" as default


eordano:

michaltakac:
`budo` as a static file server
https://github.com/mattdesl/budo/blob/master/docs/api-usage.md#http-server
For setting different default index, I guess this should do:
https://github.com/mattdesl/budo/blob/master/docs/api-usage.md#opts -> `staticOptions` -> { index: 'scene.html' }

cazala:
@michaltakac yep that's what I did, it works, but I'm having problems to serve `tmp/dcl-app` when `isDev` is `true`. It looks like it's trying to require the directory?

```js
const baseDir = isDev ? './tmp/dcl-app/' : './'
  budo(baseDir, {
    host: '0.0.0.0',
    live: true,
    port: 2044,
    stream: process.stdout,
    staticOptions: {
      index: 'scene.html'
    }
  });
```

```
Cannot find module '/Users/juancazala/Code/GitHub/cli/tmp/dcl-app'
```

It works fine if I leave the `baseDir` as `'.'`

cazala:
nevermind, this one works:

```js
  const baseDir = isDev ? './tmp/dcl-app/' : '.'
  budo('.', {
    host: '0.0.0.0',
    live: baseDir,
    dir: baseDir,
    port: 2044,
    stream: process.stdout,
    staticOptions: {
      index: 'scene.html'
    }
  });
```

PR #26, opened 2018-01-17
Serve "scene.html" as default

cazala:
Serve `scene.html` as index page, use `./tmp/dcl-app` as baseDir for development

To the reviewer:

1. to test this checkout `feature/serve-scene`
2. do `npm run build && npm start`
3. when inside the `dcl` console do `init` (enter enter enter enter)
4. then `start`. 

You should be able to open `http://localhost:2044` and get the preview of the scene served.

Issue: #14

### DIFF
--- a/src/utils/start-server.ts
+++ b/src/utils/start-server.ts
@@ -1,13 +1,18 @@
 import chalk from 'chalk';
+import isDev from './is-dev';
 const budo = require('budo');
 
 export default function (args: any, vorpal: any, callback: () => void): void {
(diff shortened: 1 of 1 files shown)


michaltakac:
I'm seeing something about `bufferutil` in console, but the scene is served as a default correctly. Did you encounter such message?
screen shot 2018-01-17 at 18 18 27

cazala:
I run it a couple times and I got that error message once.. but I don't know how to reproduce it. I'll investigate.

cazala:
Closed in favor of #28

Issue #27, opened 2018-01-17
Hot reload not working


cazala:
If you change the scene.html file while the server is running nothing happens

Issue #12, opened 2018-01-15
Fix "Error: cannot find module "(current dir)"


eordano:

michaltakac:
`budo` should serve `scene.html` from DCL project directory (where the `scene.json` resides) but right now it thinks the directory is a module. It should serve static `scene.html`.
https://github.com/mattdesl/budo/issues/103

michaltakac:
It also seems like that #14 would solve the issue, because we need to serve just the static assets.
https://github.com/mattdesl/budo/blob/master/docs/api-usage.md#http-server
https://github.com/mattdesl/budo/blob/master/docs/api-usage.md#opts -> `staticOptions`

PR #28, opened 2018-01-17
[feature] Live Server

### LINKED ISSUE #13: Fix "Error: cannot find module "(current dir)" from "(current dir)"
image

cazala:
This PR fixes #13, #14 and #27, and a bunch of bonuses:

- Fix hot-reload
- Serve "scene.html" as default
- Open browser when `serve` or `start` is invoked
- Fix a few errors shown in the console: [1](https://github.com/decentraland/cli/issues/12) and [2](https://github.com/decentraland/cli/pull/26#issuecomment-358377219)

I removed `budo` in favor of `live-server`. This is because `budo` is a development server for `browserify` that expects us to provide an entry point to bundle using browserify. That's why it was blowing up (#13) when using the CLI outside of the development environment (like with `npm link` and using it from a different project), because it was trying to find a file to browserify on.

We don't need browserify, we just need to serve static assets, watch for changes, and reload the page (this is for now, later we should hot-reload without doing a full page refresh, but for this we need to add support for that from the client).

I had to do something kinda nasty, moving the client script tag from the <head> to the very bottom of the html file, (even outside the body) this is because the client does a [document.write(...)](https://github.com/decentraland/client/blob/master/src/preview.js#L11) as soon as it's loaded and that basically kills the code injected by the live server (basically connecting via a websocket to the server so the browser knows when a file changed and reloads the page).

In order to review this PR:

1. checkout `feature/serve-scene`
2. do `npm run build && npm start`
3. inside the `dcl` console do `init` (enter enter enter enter)
4. then `start`

Your browser should open on `http://localhost:2044` and the preview of the scene should be served.


michaltakac:
Looks good, thanks!

NicoSantangelo:
Oh geez I didn't have a chance to even try it.
I'll do it on `master`

PR #29, opened 2018-01-17
Use decentraland-contracts's LANDRegistry

NicoSantangelo:
Because `decentraland-commons` now bundles the `Log` module on development ( [here](https://github.com/decentraland/commons/blob/master/src/browser.js#L2) ) we can remove the hardcoded `LANDRegistry` contract and use `decentraland-contracts` instead.

The problem is that the contract address is (and was) hardcoded. In every other project we use an ENV variable ( `LAND_REGISTRY_CONTRACT_ADDRESS` ) to determine it, checking https://github.com/decentraland/contracts#contracts-addresses which should be updated.
This problem is not introduced by this PR, but we'll need to address it soon.

Any ideas?


pages/linker.js
```diff
@@ -2,11 +2,13 @@ import "babel-polyfill";
...
 import { eth } from 'decentraland-commons';
-import LANDRegistry from '../contracts/LANDRegistry';
+import { LANDRegistry } from 'decentraland-contracts';
 
 async function ethereum() {
   await eth.connect(null, [LANDRegistry])
-  const land = await eth.getContract('LANDRegistry')
+
+  const land = LANDRegistry.getInstance()
+  land.address = '0x9519216b1d15a91e71e8cfa17cc45bcc7707e500'
```
NicoSantangelo: ^ Ropsten contract address. Keeps the implementation from https://github.com/decentraland/cli/pull/29/files#diff-519eb343357192dc49445aa44ff67531L11

michaltakac:
Since we have also a small API for this linking app, we can use dotenv there on the server and add new route like `/api/contract-address` which sends one loaded from env variable to client.

NicoSantangelo:
Yes!
I'll add that to this PR

michaltakac:
For developers, yeah.
Correct ENV with correct LAND address should be added to release and published, I think that won't  need actual CLI users attention. If that's possible :D

michaltakac:
@NicoSantangelo I see that at the bottom of the readme there is explanation how to work with `.env`, but can we have `.env.example` and tell developers to copy it to `.env` and fill out `LAND_REGISTRY_CONTRACT_ADDRESS`? Also please put `.env` to `.gitignore`.

NicoSantangelo:
Done! I skipped the `.env.example` as it was only for development, but on second thought it's better to have it there

NicoSantangelo:
Updated to use the new commons/contract API:
- https://github.com/decentraland/contracts/pull/6
- https://github.com/decentraland/commons/pull/16

needs an `npm update`

package.json
```diff
@@ -62,11 +62,13 @@
   "devDependencies": {
     "@types/chai": "^4.0.10",
     "@types/chalk": "^2.2.0",
+    "@types/fs-extra": "^5.0.0",
```
cazala: This is already under `dependencies` but I think it makes more sense to be here under `devDependencies`, so maybe we should remove the one in `dependencies`?

package.json
```diff
@@ -62,11 +62,13 @@
...
     "@types/inquirer": "0.0.36",
     "@types/koa": "^2.0.43",
     "@types/koa-router": "^7.0.27",
     "@types/koa-static": "^4.0.0",
     "@types/mocha": "^2.2.45",
+    "@types/next": "^2.4.7",
```
cazala: Same as the comment above ☝️

package.json
NicoSantangelo: Agreed, updating `dependencies`

PR #37, opened 2018-01-18
[fix] `dcl serve`

### LINKED ISSUE #36: Fix `dcl serve` in dev mode
Now that projects have docker names the command `dcl serve` fails because it was hardcoded to `./tmp/dcl-app`

cazala:
Fixes #36 and also fails gracefully if running `dcl serve` without running `dcl init` first.

To  review:

1. checkout branch `fix/serve`
2. `npm run build`
3. `npm start -- init`
4. `npm start -- serve`
5. It should open the browser and serve the scene, on the console it should say `Serving ./tmp/[project-name] at [host]:[port]`

### DIFF
--- a/src/utils/serve.ts
+++ b/src/utils/serve.ts
@@ -1,10 +1,20 @@
 import chalk from 'chalk';
 import { isDev } from '../utils/is-dev';
+import * as fs from 'fs';
 const liveServer = require('live-server');
 
 export function serve(vorpal: any, args: any, ): void {
   vorpal.log(chalk.blue('Parcel server is starting...\n'));
-  const dir = isDev ? './tmp/dcl-app' : '.';
+  let dir = '.'
+  if (isDev) {
+    try {
+      dir = `./tmp/${fs.readdirSync('./tmp')[0]}`
+    } catch (e) {
+      // this happens only in dev mode when you run `dcl serve` without running `dcl init` first
+      console.error('Project not initialized!') 
+      return
+    }
+  }
   liveServer.start({
     port: 2044,
     host: '0.0.0.0',


michaltakac:
Your implementation works great! I see that it takes first project from `tmp` directory and serves it by default. But if I created, while developing, 2 and more test projects, I cannot pick which project I want to serve. Another thing is that the checking if project I want to serve is valid DCL project (i.e. `scene.json` exists in folder I'm going to serve) is useful even outside of development.

I'm suggesting something like this:

```javascript
import chalk from 'chalk';
import { isDev } from './is-dev';
import { prompt } from './prompt';
import * as project from './project';
import * as fs from 'fs-extra';
const liveServer = require('live-server');

function runServer(vorpal: any, dir: string, callback: () => void): void {
  const isDclProject = fs.pathExistsSync(`${dir}/scene.json`);
  if (!isDclProject) {
    vorpal.log(`Seems like this is not a Decentraland project! ${chalk.grey('(\'scene.json\' not found.)')}`);
    callback();
    return;
  }

  liveServer.start({
    port: 2044,
    host: '0.0.0.0',
    root: dir,
    open: true,
    ignore: '.decentraland',
    file: 'scene.html',
    wait: 500,
    logLevel: 3
  });
}

export function serve(vorpal: any, args: any, callback: () => void) {
  vorpal.log(chalk.blue('Parcel server is starting...\n'));

  let dir = '.'
  let projectName = project.getDefaultName();

  if (isDev) {
    prompt('(Development-mode) Project name you want to upload: ', projectName)
      .then((name: string) => {
        dir = `./tmp/${name}`;
        runServer(vorpal, dir, callback);
      })
      .catch((err: Error) => {
        vorpal.log(err.message);
        callback();
        return;
      });
  } else {
    runServer(vorpal, dir, callback);
  }
}
```

This way you can pick whatever project you want to serve while in development mode, and have the check "if this folder contains valid DCL project" also available for actual users of `cli` tool.

What do you guys think?

cazala:
do we even need multiple projects while on dev mode? why don't we just create a `tmp` folder and put the project there? every time we rebuild the `tmp` folder is wiped anyway. and there will be less prompts to worry about (it would be the same experience as on non-dev mode)

cazala:
like, would it make the whole code simpler to go this way?

```js
liveServer.start({
    port: 2044,
    host: '0.0.0.0',
    root: isDev ? 'tmp' : '.',
    open: true,
    ignore: '.decentraland',
    file: 'scene.html',
    wait: 500,
    logLevel: 3
  });
```

michaltakac:
> do we even need multiple projects while on dev mode? 

Actually no, valid point, I'm fine with going with `./tmp/${fs.readdirSync('./tmp')[0]}`.

I would say this would work:
```javascript
liveServer.start({
    port: 2044,
    host: '0.0.0.0',
    root: isDev ? `./tmp/${fs.readdirSync('./tmp')[0]}` : '.',
    open: true,
    ignore: '.decentraland',
    file: 'scene.html',
    wait: 500,
    logLevel: 3
});
```

michaltakac:
Well, as we talked about this during the call, if we switch to doing the `dcl init` inside the directory you want to use for DCL project, then yes, we can get rid of all the

```javascript
if (isDev) {
    projectName = await prompt('(Development-mode) Project name you want to upload: ', projectName);
}
```
and
```javascript
 const root = isDev ? `tmp/${projectName}` : '.';
```

cazala:
yea we are changing the way `init` works so it initializes the current directory, instead of creating a new directory. So we can even get rid of the `tmp/${projectName}` to be just `'tmp'`.

Issue #42, opened 2018-01-18
Client not working with live-server


cazala:
Client not working with live-server. The injected <script> tag by the live server inside the <body> makes the client throw an error and die.

PR #43, opened 2018-01-18
fix client url, live-reload, and added watch script

### LINKED ISSUE #35: Fetch the right version of preview.[hash].js
We need to fetch `https://client.decentraland.today/assets.json` to get the latest version of `preview` and inject a script tag with it.

cazala:
Fixes #35 and #42, I also added a `watch` script that we can run after the first `npm run build` to do incremental compilations (it doesn't rebuild the linker tho, for that one we need to re-run `npm run build` after we make changes to it).

### DIFF
--- a/src/utils/generate-html.ts
+++ b/src/utils/generate-html.ts
@@ -1,26 +1,31 @@
+import axios from 'axios'
+
 interface GeneratorSettings {
   withSampleScene?: boolean;
 }
 
-export function generateHtml({ withSampleScene = false }: GeneratorSettings): string {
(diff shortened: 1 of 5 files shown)


src/utils/generate-html.ts
```diff
@@ -1,26 +1,31 @@
+import axios from 'axios'
+
 interface GeneratorSettings {
   withSampleScene?: boolean;
 }
 
-export function generateHtml({ withSampleScene = false }: GeneratorSettings): string {
+export async function generateHtml({ withSampleScene = false }: GeneratorSettings): string {
```
michaltakac: please update return type `string` -> `Promise<string>`.

Issue #10, opened 2018-01-15
`update-linker` could be an automated check (only run every 24 hours) that runs on the background


eordano:

michaltakac:
Copy-pasting a chat where we were discussing this:

**cazala** 3:49 PM
@michaltakac when does the user have to run the `upgrade` command?

**michaltakac** 4:04 PM
if there's a newer version of linker app (and I don't know when there will be a need for updating visuals of the page after github.com/decentraland/cli/issues/15 is finished)
 
so in theory, even manual upgrading is fine because it won't be much needed (contract address is automatically updated because we rely on `https://decentraland.github.io/contracts/addresses.json`)

**michaltakac** 4:15 PM
and what we can do is to add `isOutdated()` checker whenever you run `dcl {{command}}` within DCL project and if it returns `true`, it will log a message `The Ethereum linker app is outdated. Please update it by running dcl upgrade.`

michaltakac:
Closed #10 through #47.

PR #44, opened 2018-01-18
automate checking if linker client is outdated

michaltakac:
#10 
- [x] `upgrade` command for getting latest version of linker client app
- [x] check if outdated
-  ~automate upgrading~
- [x] automate checking if linker is outdated


src/utils/is-outdated.ts
```diff
@@ -0,0 +1,20 @@
+import fs = require('fs-extra');
+import { cliPath }from './cli-path';
+import { isDev } from './is-dev';
+
+export function isOutdated(vorpal: any, callback: () => void): boolean {
```
NicoSantangelo: Isn't it a bit weird that this method returns a boolean but it also controls the flow of the caller at the same time using `callback`?

what if I want to call `callback` later using the result of `isOutdated` along with some other information

michaltakac: You're right, it would be better to only focus this util for checking if it is outdated and throw away `callback` from fn args and remove this:
```javascript
try {
    fs.accessSync(`${path}/.decentraland/linker-app/_next`, fs.constants.R_OK | fs.constants.W_OK)
} catch (err) {
    vorpal.log(`Looks like linker app is missing. Try to re-initialize your project.`)
    callback()
    return
}
```
as it is kinda implemented in similar fashion in `upgrade` command already (check if it's DCL project).

michaltakac:
~automate upgrading~ ---> automate checking if linker is outdated

Vorpal will check if linker is outdated before user runs any command. If it is, only then it prints `Ethereum linker app is outdated! Please run dcl upgrade!` message.

src/utils/linker-checker.ts
```diff
@@ -0,0 +1,20 @@
+import chalk from 'chalk';
+import fs = require('fs-extra');
+import path = require('path');
+import inquirer = require('inquirer');
+import * as project from './project';
+import { cliPath }from './cli-path';
+import { isDev } from './is-dev';
+import { isOutdated } from './is-outdated';
+import { wrapAsync } from './wrap-async';
```
NicoSantangelo: Most of these imports are never used

michaltakac: lol, result of a copy-paste from another file :D removed.

PR #47, opened 2018-01-19
inject parcel boundaries based on `scene.json` data

### LINKED ISSUE #38: Parcel boundaries
Show parcel boundaries in `scene.html` while serving it with `dcl serve`. Take parcels array from `scene.json`.

https://github.com/decentraland/client/issues/15

michaltakac:
Fix #38
Blocked by https://github.com/decentraland/client/pull/17.

Injects a parcel boundary mesh to preview, based on parcel data from `scene.json`.

1. `npm run build`
2. `npm start -- init`
3. `npm start -- serve`

Production:

1. `npm link`
2. `mkdir dcl-app`
3. `cd dcl-app`
4. `dcl init`
5. `dcl serve`

### DIFF
--- a/parcel-boundary.js
+++ b/parcel-boundary.js
@@ -1,0 +1,71 @@
+document.addEventListener('DOMContentLoaded', () => {
+  const scene = window.scene;
+  if (!scene) return;
+
+  var oReq = new XMLHttpRequest();
+  oReq.onload = injectBoundaries;
+  oReq.open("get", "../scene.json", true);
+  oReq.send();
+
+  function injectBoundaries (e) {
+    var sceneMetadata = JSON.parse(this.responseText);
+    // Get the bounds of the parcel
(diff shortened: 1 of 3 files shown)


michaltakac:
The URL `/?position=0,0` is something different - it just tells you where are you standing right now and it changes after you move to different parcel within preview.

On `client`, you're also able to write `/?position=10,20` and you should spawn on parcel `10,20` - this is not working in preview (found this out right now).

The problem is that the boundary is not centered to your view, that means that if you wrote `-97,-71; -98,-72; ... ` in prrompt for `Parcels` while doing `dcl init`, the boundary would be far in the distance and you're standing at parcel `0,0` after refresh.

I'll have to add parcel centering to re-calculate where user stands, like in `editor`.

michaltakac:
@cazala added offset, boundaries should be visible now.
You need to re-init the project.

PR #57, opened 2018-01-30
[fix] IPNS links

cazala:
This PR solves the bug that is preventing two projects to have different IPNS links. The problem was that the IPFS key used to generate the IPNS hash was the Peer ID (one per IPFS node), so all the projects using the same IPFS daemon would end up having the same IPNS link.

I added a `project.json` in `.decentraland` that cointains a unique id for the project, and the IPFS key's **name** (note that this is only a reference to the private key, the actual private key is stored in the ipfs node. This hash is actually what ends up being the ipns hash in the end, so it can be public).

I tried to add `export` / `import` commands to be able to get the `.pem` and load it again (ie. from a different computer) but I ran into some weird IPFS exceptions, I'll leave that one for a second PR.

Currently the work around for the situation above would be to generate a new IPFS key and re-link.

**To the reviewer**

I already published `v0.2.9` under the `next` tag, so to test this you can just `npm i -g decentraland@next`, then create two separate projects, and push each of them. You should see they generate different IPNS hashes on the console logs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "dcl": "./bin/dcl"
   },
   "dependencies": {
+    "@types/uuid": "^3.4.3",
     "axios": "^0.17.1",
     "babel-polyfill": "^6.26.0",
     "chalk": "^2.3.0",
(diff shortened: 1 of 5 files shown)


cazala:
`decentraland@0.2.9` moved from `next` to `latest`

PR #75, opened 2018-03-01
Connect with Decentraland IPFS node

### LINKED ISSUE #66: Pin scenes to decentraland ipfs node
## Motivation

As we are developing an own Decentraland IPFS node, we could pin created/updated scenes.

With this, we are not going to be dependent on the local user IPFS node to be up when getting the scene from our apps.

## Concept Implementation

I've worked in an [experimental branch](https://github.com/decentraland/cli/tree/connect-with-ipfs-node), only focused on functional things.

### We need to: 

- store the peerId of the local user IPFS node as [here](https://github.com/decentraland/cli/blob/connect-with-ipfs-node/src/utils/uploader.ts#L69)

- pin the files at Decentraland IPFS node once [update land data transaction is completed] (https://github.com/decentraland/cli/blob/connect-with-ipfs-node/pages/linker.js#L164) to be always available. [And when an update occurs](https://github.com/decentraland/cli/blob/connect-with-ipfs-node/src/utils/uploader.ts#L111)

- show the process [flow status](https://github.com/decentraland/cli/blob/connect-with-ipfs-node/pages/linker.js#L214) to the user (fully experimental)

### Endpoint (@ALPHA stage) 

```
@GET http://ipfs.decentraland.zone:3000/api/pin/:peerid/:x/:y

Response: { ok: 'true' } or {'ok': false}
```
- Needs https
- We can make it more verbose to handle errors better.

### Test

`npm start -- init`

Fill at least the parcels you own

#### Pin new scene

`npm start -- upload`

`npm start -- link`

#### Update Scene

 Make a change in your scene, then run:

`npm start -- upload`

To see them reflected you can use:

http://ipfs.decentraland.zone:3000/api/resolve/:ipns?dependencies=true

E.g: http://ipfs.decentraland.zone:3000/api/resolve/QmTZZnwcsUJuQL2EH3oLZBaA6HZvhv8KJ7Qd9QP9LpEjW1?dependencies=true

Thanks @menduz for the early catch.

nachomazzara:
Resolve #66


pages/linker.js
```diff
@@ -129,8 +156,8 @@ export default class Page extends React.Component {
       try {
         console.log('update land data', coordinates, data)
         const tx = await land.updateManyLandData(coordinates, data)
-        this.setState({ tx })
-        closeServer(true, 'transaction successful')
+        this.watchTransactions(tx, coordinates[0].x, coordinates[0].y)
```
menduz: does this watch the transaction for the whole Estate?

pages/linker.js
```diff
@@ -141,6 +168,22 @@ export default class Page extends React.Component {
     }
   }
 
+  async watchTransactions (txId, x, y) {
+    const {peerId} = this.state
+    const tx = await txUtils.waitForCompletion(txId)
+    if (!txUtils.isFailure(tx)) {
+      this.setState({transactionLoading: false, pinningLoading: true})
```
menduz: Format code before push

pages/linker.js
```diff
@@ -155,6 +198,22 @@ export default class Page extends React.Component {
...
+  renderTransactionStatus = () => (
+    this.state.tx ?
+      !this.state.transactionLoading ?
+        <p style={{ color: 'green' }}>{`Transaction confirmed.`}</p>
+        :  <p style={{ color: 'orange' }}>{`Transaction pending. Will take a while...`}</p>
+      : null
+  )
+
+  renderPinningIPFSStatus = () => (
+    !this.state.error && this.state.tx && !this.state.transactionLoading ?
```
menduz:
```ts
if(..) {
  return
} else if(..) {
  return
}
return null
```
Is probably easier to read

pages/linker.js
nachomazzara: Wait until transaction is [completed by success or fail](https://github.com/decentraland/commons/blob/master/src/ethereum/txUtils.js#L48)

PR #90, opened 2018-03-13
Add Segment.io

eordano:


src/utils/analytics.ts
```diff
@@ -0,0 +1,81 @@
...
+/// List of tracked events
+///
+
+export const sceneCreated = track('Scene created');
+export const preview = track('Preview started');
+export const sceneUploadRequested = track('Scene upload started');
+export const sceneUploaded = track('Scene uploaded');
+export const sceneLink = track('Scene ethereum link started');
+export const sceneLinked = track('Scene ethereum linked');
+export const deploy = track('Scene deploy requested');
```
cazala: what's the difference between the `Scene deploy requested` and the `Scene upload started` events?

src/utils/uploader.ts
```diff
@@ -105,6 +109,9 @@ export async function uploader(vorpal: any, args: any, callback: () => void) {
     vorpal.log('Updating IPNS reference to folder hash... (this might take a while)');
     const { name } = await ipfsApi.name.publish(ipfsHash, { key: project.id });
 
+    // Upload successful
+    sceneUploaded();
```
matiasbargas: have sense to send the ipns name to the track properties?

matiasbargas: same with scenes that are being linked

src/utils/analytics.ts
eordano: `push` vs `upload` -- hmm maybe not necessary I'll work on simplifying this

src/utils/uploader.ts
eordano: Roger

PR #96, opened 2018-03-14
Load files based on .dclignore & check main format

### LINKED ISSUE #95: Define conventions for folder structure and warnings
We need to have a specific folder which contents can be fully uploaded to IPFS.

We need clearly display warnings for folders that are ignored., or, we need to define something similar to the npm/git ignore behaviour.

nachomazzara:
# Context

- Add `.decentralandignore` at the `root`
- Upload all the files inside the `root` folder if they are not ignored in `.dclignore`
- Check if the main file in `scene.json` is a supported format

# Tickets

Fix #95

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -112,26 +112,15 @@ export function init(vorpal: any) {
         fs.outputFileSync(
           path.join(dirName, '.dclignore'),
           [
-            '.gitgnore',
-            '.decentraland',
-            '.npmignore',
-            '.editorconfig',
-            '.vs/',
-            '.vscode/',
-            '.idea',
-            '.babelrc',
-            '.env',
+            '.*',
             'package.json',
             'package-lock.json',
             'yarn-lock.json',
-            'node_modules/',
-            'bower_components/',
-            'README.md',
-            '*.log',
-            'dist',
             'build.json',
             'tsconfig.json',
-            'tslint.json'
+            'tslint.json',
+            'node_modules/',
+            'dist/'
           ].join('\n')
         );
 
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -110,7 +110,7 @@ export function init(vorpal: any) {
         );
 
         fs.outputFileSync(
-          path.join(dirName, '.decentralandignore'),
+          path.join(dirName, '.dclignore'),
           [
             '.gitgnore',
             '.decentraland',
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -109,6 +109,32 @@ export function init(vorpal: any) {
           )
         );
 
+        fs.outputFileSync(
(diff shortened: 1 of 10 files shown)


src/utils/uploader.ts
```diff
@@ -19,34 +20,50 @@ export async function uploader(vorpal: any, args: any, callback: () => void) {
...
 
-  const isDclProject = await fs.pathExists(path.join(root, 'scene.json'));
-  if (!isDclProject) {
+  try {
+    const scene = JSON.parse(fs.readFileSync(path.join(root, 'scene.json'), 'utf-8'));
+    const isWebSocket = /(ws(s?))\:\/\//gi;
+    const supportedExtensions = ['js', 'html', 'xml'];
+    const mainExt = scene.main.split('.').pop();
+    if (
+      !isWebSocket.test(scene.main) &&
```
eordano: If you take this approach:
```
const isWebSocket = (str: string) => /(ws(s?))\:\/\//gi.test(str)
const invalidFormat = (str: string) => !(supportedExtensions.filter(ext => ext === mainExt).length)
const mainExists = fs.existsSync(path.join(root, scene.main))
```
 then the `if` clause is way way more readable:

```
if (!isWebSocket(main)) {
  if (invalidFormat) return exitWithError(`Main scene format file (${scene.main}) is not a supported format`)
  if (!mainExists) return exitWithError(`Main scene format file ${scene.main} is missing`)
}
```

src/commands/init.ts
```diff
@@ -109,6 +109,32 @@ export function init(vorpal: any) {
           )
         );
 
+        fs.outputFileSync(
+          path.join(dirName, '.decentralandignore'),
+          [
+            '.gitgnore',
```
eordano: Let's do `.*` to start.
Let's drop `bower_components`
Let's just leave things that are generated by this same file.

src/utils/uploader.ts
```diff
@@ -7,46 +7,66 @@ import { getIPFSURL } from './get-ipfs-url';
...
+  try {
+    const scene = JSON.parse(fs.readFileSync(path.join(root, 'scene.json'), 'utf-8'));
+    const supportedExtensions = ['js', 'html', 'xml'];
+    const mainExt = scene.main.split('.').pop();
+    const isWebSocket = (str: string) => /(ws(s?))\:\/\//gi.test(str);
+    const isInvalidFormat = !supportedExtensions.filter(ext => ext === mainExt).length;
+    const mainExists = fs.existsSync(path.join(root, scene.main));
+
+    if (!isWebSocket(scene.main)) {
+      if (isInvalidFormat) exitWithError(`Main scene format file (${scene.main}) is not a supported format`);
```
eordano: Please use `return` so execution doesn't continue
Also use brackets and enter :) my previous example was as a demostration

PR #114, opened 2018-03-21
Handle & show ipfs-node pinning errors

nachomazzara:
- Move pinning to utils
- Handle & show error if `ipfs-node` return with error
- Only tracking `link success` when there are no errors
- Close server when pinning finished


src/utils/pin-files.ts
```diff
@@ -0,0 +1,16 @@
+import axios from 'axios';
+import { getIPFSURL } from './get-ipfs-url';
+
+export const pinFiles = async (peerId: string, { x, y }: any) => {
+  const ipfsURL: string = await getIPFSURL();
+
+  return axios
+    .post(`${ipfsURL}/pin/${peerId}/${x}/${y}`)
```
menduz: Do not mix async code and `then` notation

src/utils/uploader.ts
```diff
@@ -145,35 +144,16 @@ export async function uploader(vorpal: any, args: any, callback: () => void) {
...
+      await pinFiles(project.peerId, { x, y });
       // Upload successful
-      sceneUploadSuccess({ ipfsURL, ipfsHash, ipnsHash });
-
-      vorpal.log(`Pinning files ${ok ? 'success' : 'failed'}`);
+      sceneUploadSuccess({ ipfsHash, ipnsHash });
+      vorpal.log('Pinning files to IPFS succeeded');
     } catch (err) {
-      vorpal.log('Error', JSON.stringify(err.message));
+      vorpal.log('Pinning files to IPFS failed: ', JSON.stringify(err.message));
```
cazala: `err.message` is already a string, no need to stringify

PR #131, opened 2018-04-11
test: added integration tests for init

belohlavek:
Added integration tests for the init method.
Also made sure we can run stuff on an specific directory.


test/integration/sandbox.ts
```diff
@@ -0,0 +1,17 @@
+import * as fs from 'fs-extra'
+import * as path from 'path'
+const rimraf = require('rimraf')
```
nachomazzara: Could we change it to `import * as rimraf from 'rimraf`

PR #134, opened 2018-04-12
fix: getFiles and getAllFilePaths

belohlavek:
Fixed and tested getFiles and getAllFilePaths.
The issue was that we weren't uploading nested folders, that was due to several small issues which are fixed with this pr.


belohlavek:
I promise @nachomazzara reviewed this one 😂

PR #135, opened 2018-04-13
fix: websocket example and metaverse-api

belohlavek:
Closes [CLIENT-77](https://decentraland.atlassian.net/browse/CLIENT-77)

What?

We need to fix the main field in the scene.json for multiplayer projects.

Why?

It fails to create multiplayer projects.

Steps to reproduce:

dcl init (choose multiplayer)dcl preview


src/lib/Preview.ts
```diff
@@ -21,16 +21,16 @@ export class Preview extends EventEmitter {
...
           </head>
           <body>
           </body>
         </html>
       `)
     })
 
-    // serve the folder `dcl-sdk/artifacts` as `/dcl-sdk`
-    this.app.use('/dcl-sdk', express.static(path.dirname(require.resolve('dcl-sdk/artifacts/preview'))))
+    // serve the folder `metaverse-api/artifacts` as `/metaverse-api`
```
nachomazzara: Do you still think the comment is necessary? If so, we are using only `preview` as `/metaverse-api`

PR #144, opened 2018-04-19
fix: resolve promise even when tracking disabled

michaltakac:
If user disabled sending usage info, commands should work (Promise now resolves).


belohlavek:
waiting for permissions to merge pr

PR #147, opened 2018-04-23
fix: preview must work offline

belohlavek:


src/utils/moduleHelpers.ts
```diff
@@ -31,6 +31,9 @@ export function buildTypescript(): Promise<void> {
 }
 
 export async function latestVersion(name: string): Promise<string> {
+  if (!await isOnline()) {
```
nachomazzara: We can store the tracking and send a bash with them once the user is online again

PR #174, opened 2018-06-23
feat: better lifecycle for linker app

eordano:
This pull request:

* Decouples the interaction and reliance on the browser window to control the lifecycle of the server
  - Do not close the server on the first error (if metamask is locked, the server exits immediately)
  - Do not close the tab on success (it's ok to have the tab open after success)
* Removes an error when using `dcl` in a docker vm where `open` is not available
* Minor TODO added to `Ethereum.ts`

### DIFF
--- a/linker-app/src/index.tsx
+++ b/linker-app/src/index.tsx
@@ -34,7 +34,7 @@ async function getIPFSKey() {
 }
 
 async function closeServer(ok, message) {
-  console.log('closing server:', message)
+  console.log('sending error to server:', message)
   await fetch(`/api/close?ok=${ok}&reason=${message}`)
 }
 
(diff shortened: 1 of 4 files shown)


belohlavek:
I would go about this in a different way:

* Fatal errors should still cause a reject() on the linker api's side (allows a CI to fail instead of staying idle).
* Fatal errors should be more granular
* Browser errors should be replicated in the terminal (emitted) for more visibility
* Specific errors (locked wallet, etc) should be handled gracefully with separate strategies on the dApp side
* Logic should work for both link and deploy command

I agree that the linker window should stay open.

eordano:
Closing it for the time being until a better analysis is made

belohlavek:
On another note, I had my head somewhere else: the linker app won't ever be opened by the CI. My other points still stand.

PR #180, opened 2018-07-04
fix: make same validations as dcl deploy for dcl link

fmiras:


src/lib/Decentraland.ts
```diff
@@ -176,4 +176,10 @@ export class Decentraland extends EventEmitter {
   private pipeEvents(event: string, ...args: any[]) {
     this.emit(event, ...args)
   }
+
+  private async validateOwnership() {
```
belohlavek: we should put this in Ethereum

PR #181, opened 2018-07-05
fix: web3 check ownership on linker-app

fmiras:
Added ownership check on Linker dApp


belohlavek:
Fede is unstoppable 😂

PR #184, opened 2018-07-12
 feat: show test message if in development mode

fmiras:
With this a message is going to show up if `DCL_ENV` envrioment variable is turned on 'dev'

screen shot 2018-07-12 at 11 50 29

### DIFF
--- a/linker-app/src/index.tsx
+++ b/linker-app/src/index.tsx
@@ -2,7 +2,7 @@ import * as React from 'react'
 import * as ReactDOM from 'react-dom'
 
 import { txUtils } from 'decentraland-eth'
-import { Address, Blockie, Header, Navbar, Menu, Button, Radio, Loader } from 'decentraland-ui'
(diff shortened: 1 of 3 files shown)


linker-app/src/index.tsx
```diff
@@ -18,7 +18,7 @@ interface IOptions {
   base: boolean
 }
 
-interface IState {
+interface State {
```
belohlavek: IState is correct, follows conventions

linker-app/src/index.tsx
```diff
@@ -162,16 +176,34 @@ export default class Page extends React.Component<any, IState> {
           </React.Fragment>
         )}
         <style>{`
-          body {
+          .LinkScenePage {
```
belohlavek: why uppercase?

linker-app/src/index.tsx
fmiras: I wanted to follow same conventions as showing on our SDK docs https://docs.decentraland.org/sdk-reference/scriptable-scene what do you think of that?

PR #185, opened 2018-07-12
chore: move static scene project option to the end

fmiras:
screen shot 2018-07-12 at 11 58 10


belohlavek:
Holding on to this one as it will go as a hotfix

PR #189, opened 2018-07-13
chore: track analytics rejections

belohlavek:


menduz:
We are using single user because we use a free version of Segment. That's the reason we track the devId as a "tag" or "label"

PR #192, opened 2018-07-17
feat: new basic typescript boilerplate

fmiras:


src/samples/ts-static/scene.tsx
```diff
@@ -0,0 +1,14 @@
+import { createElement, ScriptableScene } from 'metaverse-api'
+
+export default class RollerCoaster extends ScriptableScene<any, { time: number }> {
```
menduz: Please remove `<any, { time: number }>` since this scene doesn't use state nor props. The focus of this example is to put the fewer things possible in the scene to reduce the visual stress of non-developer persons

src/samples/ts-static/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "description": "My new Decentraland project",
+  "scripts": {
+    "build": "metaverse-compiler build.json",
+    "watch": "metaverse-compiler build.json --watch"
+  },
+  "author": "",
+  "license": "MIT",
+  "devDependencies": {
+    "metaverse-api": "^4.0.0",
+    "typescript": "^2.7.2"
```
menduz: Please remove typescript from dependencies, the metaverse-api already contains the correct version of typescript

PR #214, opened 2018-08-27
fix: hot reloading wasn't working

fmiras:


menduz:
The things like 
```ts
shouldWatch = (args.options.watch !== undefined ? args.options.watch : true)
```
Are more readable in the shape
```ts
function getOrElse(value, default) { return value !== undefined ? value : default }


shouldWatch = getOrElse(args.options.watch, true)
```

PR #216, opened 2018-09-03
feat: make `now` work in default scenes

eordano:
Make `dcl init` provide you with a `now`-compatible project by default.

### DIFF
--- a/src/samples/basic-static/package.json
+++ b/src/samples/basic-static/package.json
@@ -5,7 +5,7 @@
   "author": "",
   "license": "MIT",
   "scripts": {
-    "start": "dcl start"
+    "start": "dcl start --ci"
   },
   "devDependencies": {
     "decentraland-api": "latest"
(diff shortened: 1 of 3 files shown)


fmiras:
I see a couple of things here:

- For having a `now`-compatible project by default we also have to install `decentraland` (cli) npm package as project local dependency, otherwise now will throw a `sh: dcl: not found` error. 

- We also should ignore `.decentraland` folder, because by default there is not `.gitignore` and since now expose your code and make it public, we should be using `now` ignore properties to not upload ipfs key by default (just because privateness)

What if we use a flag like `dcl init --now` so non-cloud users won't be afected with this 2 more changes and now users are now-ready from scratch?

belohlavek:
Adding up to what my homies mentioned above, we shouldn't be adding now specifc commands or flags. We should discuss generating a ready-to-deploy artifact. Eveything already transpiled and the start command ready to run in `--ci mode`

fmiras:
Implemented on #218 with a different approach

PR #219, opened 2018-09-05
refactor: let's have a single source of addresses and providers

fmiras:
# What? 
We aren't having a single source of truth when we talk about contract addresses and providers, linker-app was using hardcoded infura same as CLI and some functions where fetching contract addresses from [contracts.decentraland.org/addresses.json](https://contracts.decentraland.org/addresses.json) but others have also hardcoded addresses

# Why? 
Avoid a properties mess, prior env variables, otherwise fetch missing ones.

### DIFF
--- a/src/lib/Ethereum.ts
+++ b/src/lib/Ethereum.ts
@@ -27,17 +27,18 @@ export interface ILandData {
  * ethereum:get-ipns-success - Successfully fetched and parsed landData
  */
 export class Ethereum extends EventEmitter {
+  private static addresses: any
   private static contracts = new Map<string, Contract>()
 
   static async getContractAddresses(): Promise<any> {
-    if (this.contracts) {
-      return this.contracts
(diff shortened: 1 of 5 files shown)


src/lib/Ethereum.ts
```diff
@@ -28,67 +27,56 @@ export interface ILandData {
...
-      fail(ErrorType.ETHEREUM_ERROR, `Unable to fetch mana contract: ${error.message}`)
-    }
+    return envContract || this.getContractAddress('MANAToken')
   }
 
   async getLandOf(address: string): Promise<ICoords[]> {
     let res
     try {
-      const contract = await landContract
+      const contract = await Ethereum.getContract('LANDProxy')
```
belohlavek: This is the only method that calls getContract inside try catch.
I don't recall correctly which were all the causes why getContract() might fail, but let's keep it consistent across the codebase

PR #235, opened 2018-11-06
chore: update ECS example

menduz:
It fixes the example to use the latest version of the API and also adds a necessary configuration for VSCode to find the types


src/samples/ecs/.vscode/settings.json
```diff
@@ -0,0 +1,16 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib",
```
fmiras: awesome

PR #278, opened 2018-11-16
feat: create private-key flag

fmiras:
# What? 
Now developers are able to use a private key command flag to avoid the linker app step of signing to upload content of LAND, e.g.: 
```bash
dcl deploy --private-key d23c7f31d16256091a4c362351a8193eb68189fd42bfd4c2b828b17306ace4c4
```

# Why? 
With an update operator address and the `--skip` flag, now we're ready to integrate `dcl deploy` to upload content to mainnet content services through continuous integration platforms

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -848,6 +848,11 @@
         "acorn": "^5.0.0"
       }
     },
+    "aes-js": {
+      "version": "3.0.0",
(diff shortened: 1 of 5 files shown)


fmiras:
@eordano already changed to env variable! can you see any other security issue?

PR #296, opened 2018-12-06
 feat: support now by default with docker

fmiras:
# What? 
Prepare ECS boilerplate to work with ZEIT Now out of the box

# Why? 
So developers and users can easily share private decentraland experiences


src/samples/ecs/now.json
```diff
@@ -0,0 +1,4 @@
+{
+  "version": 1,
+  "type": "docker"
+}
```
menduz: Can we not add this file and create a package.json#script for now publishing?

fmiras: Done!

PR #297, opened 2018-12-12
Allow partial deploy

m-sossich:
# What? 
Deploy the only missing content in the content server (unless the full deploy is forced)
# Why? 
Avoid deploying the same content over and over again

### DIFF
--- a/src/lib/content/ContentClient.ts
+++ b/src/lib/content/ContentClient.ts
@@ -4,6 +4,7 @@ import { ContentUploadRequest } from './ContentUploadRequest'
 import { Coords } from '../../utils/coordinateHelpers'
 
 import * as request from 'request'
+import * as fetch from 'isomorphic-fetch'
 
 export type ParcelInformation = {
   parcel_id: string
@@ -86,4 +87,21 @@ export class ContentClient {
       })
     })
   }
+
+  async checkContentStatus(cids: string[]): Promise<Response> {
+    return fetch(`${this.contentServerUrl}/content/status`,{
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({ "content": cids })
+    }).then(async function(response) {
+      if (response.status >= 400) {
+        const msg = await response.json()
+        throw new Error(`Bad response from server: ${msg.error}`)
+      }
+      return response.json()
+    })
+  }
+
 }
--- a/src/lib/content/ContentClient.ts
+++ b/src/lib/content/ContentClient.ts
@@ -89,19 +89,22 @@ export class ContentClient {
   }
 
   async checkContentStatus(cids: string[]): Promise<Response> {
-    return fetch(`${this.contentServerUrl}/content/status`,{
-      method: 'POST',
-      headers: {
-        'Content-Type': 'application/json'
-      },
-      body: JSON.stringify({ "content": cids })
-    }).then(async function(response) {
+    try {
+      const response = await fetch(`${this.contentServerUrl}/content/status`,{
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json'
+        },
+        body: JSON.stringify({ "content": cids })
+      })
       if (response.status >= 400) {
         const msg = await response.json()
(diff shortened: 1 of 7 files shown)


eordano:
I love this <3

src/lib/content/ContentClient.ts
```diff
@@ -86,4 +87,21 @@ export class ContentClient {
...
+    return fetch(`${this.contentServerUrl}/content/status`,{
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({ "content": cids })
+    }).then(async function(response) {
+      if (response.status >= 400) {
+        const msg = await response.json()
+        throw new Error(`Bad response from server: ${msg.error}`)
```
fmiras: Please use `fail()` method which have analytics control + vorpal behaviour

src/lib/content/ContentClient.ts
```diff
@@ -86,4 +87,21 @@ export class ContentClient {
       })
     })
   }
+
+  async checkContentStatus(cids: string[]): Promise<Response> {
```
fmiras: Why are you using `async` here if you are manually handling promises? I think we should use all `async/await` approach to keep consistency

src/lib/content/ContentClient.ts
```diff
@@ -86,4 +87,21 @@ export class ContentClient {
...
+    return fetch(`${this.contentServerUrl}/content/status`,{
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({ "content": cids })
+    }).then(async function(response) {
```
fmiras: If we're not needing specific `this` context let's use arrow functions to keep consistency

src/lib/content/ContentClient.ts
fmiras: Can we type the `Response` object?

src/lib/content/ContentService.ts
```diff
@@ -77,4 +83,14 @@ export class ContentService extends EventEmitter {
     validity.setMonth(validity.getMonth() + 6)
     return { value: rootCID, signature: signature, pubKey: address, validityType: 0, validity: validity, sequence: 2 }
   }
+
+  private async filterUploadedContent(files: IFile[], manifest: ContentIdentifier[]): Promise<IFile[]> {
+    const cidMaps = manifest.reduce((map, obj) => (map[obj.name] = obj.cid, map), {})
```
fmiras: what does this reduce callback function exactly returns? If we're making a map of CIDs i suggest using a `Map<>` to ensure not having duplicates and also we should use the CID as key instead of the name, and have the name as value, is more accurate.

src/lib/content/ContentService.ts
```diff
@@ -77,4 +83,14 @@ export class ContentService extends EventEmitter {
...
+    const res = await this.client.checkContentStatus(Object.values(cidMaps))
+    return files.filter(f => {
+      const cid = cidMaps[f.path]
+      const uploaded = res[cid]
+      return uploaded !== undefined && !uploaded
```
fmiras: What do you want to achieve with this statements? Why don't you just return `!!uploaded` or `!!res[cid]`?

src/lib/content/ContentClient.ts
menduz: good point

src/lib/content/ContentService.ts
menduz: there can be several duplicated files (by content) with different names, those resolve to the same CID but with a diffferent file name, that is the reason of the CID as value and not as a key

src/lib/content/ContentClient.ts
m-sossich: Done

src/lib/content/ContentService.ts
m-sossich: What he said...

src/lib/content/ContentService.ts
m-sossich: Good to know. Done

PR #298, opened 2018-12-12
feat: show error message when not update authorized on linker

fmiras:
# What? 
Validates if the provided wallet address is authorized to update the base parcel of the scene.

# Why? 
Because users could have different wallets/providers than the one that the CLI use and could cause a bad upload attempt.


linker-app/src/components/LinkerPage/LinkerPage.tsx
```diff
@@ -17,53 +17,84 @@ export default class LinkScenePage extends React.PureComponent<LinkerPageProps,
...
+      return (
+        <React.Fragment>
+          <p>
+            Using {this.getWalletType()} address: &nbsp;
+            <Blockie scale={3} seed={wallet.address}>
+              <Address tooltip strong value={wallet.address} />
+            </Blockie>
+          </p>
+          {'isUpdateAuthorized' in base && !base.isUpdateAuthorized ? (
+            <Error>This address doesn't either own this LAND or have update permissions.</Error>
```
nearnshaw: The Ethereum address doesn't own the LAND token and it doesn't have update permissions on it.

PR #306, opened 2018-12-26
feat: Skip owner check to allow Estate deployment

eordano:
This is a quickfix before adding Estate support.


fmiras:
@eordano Please remove the unused import of `base` variable so build pass and we can merge it :)

PR #309, opened 2018-12-26
Full deploy as default

m-sossich:
# What? 
Make the "full deployment" the default option

# Why?
Avoid confusions.

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -14,7 +14,7 @@ export interface IDeployArguments {
     host?: string
     skip?: boolean
     https?: boolean
-    force?: boolean
+    partial?: boolean
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/commands/deploy.ts
```diff
@@ -25,13 +25,13 @@ export function deploy(vorpal: any) {
...
     .option('-hs, --https', 'Use self-signed localhost certificate to use HTTPs at linking app (required for ledger users)')
-    .option('-f, --force', 'deploy the full content')
+    .option('-p, --partial', 'deploy only new content')
     .action(
       wrapCommand(async (args: IDeployArguments) => {
         const dcl = new Decentraland({
           isHttps: !!args.options.https,
           contentServerUrl: args.options.host || 'https://content-service.decentraland.zone',
-          forceDeploy: args.options.force
+          forceDeploy: !(!!args.options.partial)
```
fmiras: With one `!` we're ok :)
```ts
forceDeploy: !args.options.partial
```

m-sossich: Done

PR #318, opened 2019-01-03
chore: replace old CircleCI configuration

fmiras:
# What? 
Change `config.yml` file

# Why? 
Current config is running some commands that are only useful in old project structure + separating concerns into jobs is better for task dependencies.


fmiras:
Fixed from cloud config.

Issue #304, opened 2018-12-24
There is no explicit way to remove a deployed scene, or to move or subdivide a deployed multiparcel scene


cfravel:
--- environment
Chrome, Windows 10, no HMD
npm i -g decentraland@1.4.0-20181217174426.commit-7fe2565

--- Steps to reproduce

Deploy a valid 2-parcel scene (Scene A)
// success
// and it looks right in alpha client

Deploy a 1-parcel scene to just one of the two parcels of the first scene (Scene B)
// expected error message: Cannot deploy a scene that overlaps with part of a previously deployed scene
// actual error message: Error: This address doesn't either own this LAND or have update permissions.

Decide to adjust the 2 parcel scene in order to effectively subdivide the two parcels apart.
//There isn't any explicit way to do this per the docs or UI.

---- Partial workaround if the base parcel is the same.
Deploy a 1 parcel scene (Scene C) to the BASE PARCEL of the multiparcel Scene A
// success
Try again to deploy the 1-parcel Scene B to its single parcel (that was a non-base parcel of Scene A)
// success

--- HOWEVER
This does NOT actually delete the base parcel, and it stands in the way of any new scene being uploaded to a different base parcel.
So, for example, if you want to move a scene south or west, the base parcel of the original scene, even if reduced to a 1-parcel empty scene, will keep the new scene from being uploaded
Example
Deploy a 4 parcel scene whose base parcel is its SW corner, at n,m
"Erase" it by uploading a 1 parcel scene at n,m
Redeploy the 4 parcel scene at n-1, m-1
// fails

fmiras:
Removing a scene isn't supported. In the other hand, the CLI overrides any scene and also it know prompts a y/n question for that (#328)

PR #328, opened 2019-01-04
feat: prompt y/n question on scene overlap

fmiras:
Related #304 

# What? 
Prompt a Yes/No message if the scene to deploy is overlaping an existing one

# Why? 
Warn the developer

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -48,10 +48,10 @@ export async function main() {
 
   const dcl = new Decentraland({
     isHttps: args['--https'],
-    contentServerUrl:
-      args['--host'] || 'https://content.decentraland.today',
+    contentServerUrl: args['--host'] || 'https://content.decentraland.today',
     workingDir: args._[2],
-    forceDeploy: !args['--partial']
+    forceDeploy: !args['--partial'],
+    yes: args['--skip']
   })
 
   let ignoreFile = await dcl.project.getDCLIgnore()
--- a/src/lib/Decentraland.ts
+++ b/src/lib/Decentraland.ts
@@ -1,12 +1,13 @@
 import { EventEmitter } from 'events'
-import * as events from 'wildcards'
 import { ethers } from 'ethers'
+import * as events from 'wildcards'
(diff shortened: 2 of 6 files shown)


src/lib/Decentraland.ts
```diff
@@ -275,6 +278,29 @@ export class Decentraland extends EventEmitter {
...
+    if (
+      newScene.scene.base !== oldScene.scene.base ||
+      newScene.scene.parcels !== oldScene.scene.parcels
+    ) {
+      const results = await inquirer.prompt({
+        type: 'confirm',
+        name: 'continue',
+        default: true,
+        message:
+          'This new scene.json is overlaping an already existing scene. Do you want to continue?'
```
matiasbargas: ```The scene.json file lists parcels that overlap with an existing scene. Do you wish to overwrite the other scene?```
thanks @nearnshaw

src/commands/deploy.ts
```diff
@@ -48,10 +48,10 @@ export async function main() {
 
   const dcl = new Decentraland({
     isHttps: args['--https'],
-    contentServerUrl:
-      args['--host'] || 'https://content.decentraland.today',
+    contentServerUrl: args['--host'] || 'https://content.decentraland.today',
     workingDir: args._[2],
-    forceDeploy: !args['--partial']
+    forceDeploy: !args['--partial'],
+    yes: args['--skip']
```
michaltakac: Can you use a more descriptive key like `skipOverwrite`?

src/lib/Decentraland.ts
```diff
@@ -26,6 +26,7 @@ export type DecentralandArguments = {
   blockchain?: boolean
   contentServerUrl?: string
   forceDeploy?: boolean
+  yes?: boolean
```
michaltakac: Can you use a more descriptive key like `skipOverwrite`?

src/commands/deploy.ts
fmiras: good catch! I've already opened another PR related to that (because that flag was there before and I wanted to separate concerns) 

PR #329

PR #329, opened 2019-01-07
 feat: change skip flag name to yes

fmiras:
# What? 
Change flag name `--skip` to `--yes` or `-y`

# Why? 
It's more intuitive than `--skip` (one could think that is skipping a process like deploying itself) and adopting similar approach as for example `npm` or `apt-get`

**NOTE:** first need to merge #328

### DIFF
--- a/linker-app/src/components/LinkerPage/LinkerPage.tsx
+++ b/linker-app/src/components/LinkerPage/LinkerPage.tsx
@@ -5,7 +5,7 @@ import Navbar from 'decentraland-dapps/dist/containers/Navbar'
 import { baseParcel, isDevelopment, rootCID } from '../../config'
 import Error from '../Error'
 import { LinkerPageProps } from './types'
-import { coordsToString } from 'src/modules/land/utils'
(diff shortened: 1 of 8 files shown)


michaltakac:
Well, since users will learn about various flags through help or docs, I don't think there's a difference between `skip` and `--yes`, because I also don't know what `yes` means ("Say yes to all defaults?").

What about using flag `--o` and longer form `--overwrite` to let cli overwrite overlaps without asking, and if such flag is not provided, always ask?

fmiras:
> Well, since users will learn about various flags through help or docs, I don't think there's a difference between `--skip` and `--yes`, because I also don't know what `yes` means ("Say yes to all defaults?").
> 
> What about using flag `--o` and longer form `--overwrite` to let cli overwrite overlaps without asking, and if such flag is not provided, always ask?

The thing is that the `--yes` flag accomplish the same thing as `npm` or `apt-get` in this case, that is saying yes to all defaults (In this case is yes to size warning and yes to overlaping a scene).

But also we describe what means the `--yes` flag by executing `dcl help deploy`

PR #340, opened 2019-01-10
[WIP] fix: analytics identify

### LINKED ISSUE #343: Segment analytics identify not working properly
# What? 
There are some cases where Segment receives the identify event and not the SEND_DATA_ANON info which is right next to the identify line 

# Why? 
It may be because of the async approach of analytics

fmiras:
Fixes #343 

# What? 
On analytics identify only the first request is reaching segment

# Why? 
The `async` is not being resolved


fmiras:
Invalid issue

PR #342, opened 2019-01-10
fix: do not throw error on no previous Scene to validate

m-sossich:

### DIFF
--- a/src/lib/Decentraland.ts
+++ b/src/lib/Decentraland.ts
@@ -280,9 +280,9 @@ export class Decentraland extends EventEmitter {
       getObject(newScene.scene.base)
     )
     if (
-      oldScene != null &&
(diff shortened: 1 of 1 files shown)


src/lib/Decentraland.ts
```diff
@@ -280,8 +280,9 @@ export class Decentraland extends EventEmitter {
       getObject(newScene.scene.base)
     )
     if (
-      newScene.scene.base !== oldScene.scene.base ||
-      newScene.scene.parcels !== oldScene.scene.parcels
+      oldScene != null &&
```
fmiras: Let's use `===` or `!==` as much as we can please (JavaScript craziness)

fmiras: changed b756d22

Issue #336, opened 2019-01-08
Handle `dcl info` without target


fmiras:
# What? 
Handle error of non-target of `dcl info`

# Why? 
Improvement on DX

PR #346, opened 2019-01-10
fix: `dcl info` for negative X coords

### LINKED ISSUE #345: `dcl info` with negative X coords not working
# What? 
When executing for example `dcl info -12,-40` the CLI throw an error because it only parse the first digit as a flag 

# Why? 
Because we're not handling `arg` flag parsing

fmiras:
Fixes #345 and #336

### DIFF
--- a/src/commands/info.ts
+++ b/src/commands/info.ts
@@ -37,12 +37,41 @@ export const help = () => `
       ${chalk.green('$ dcl info 0x8bed95d830475691c10281f1fea2c0a0fe51304b')}
 `
 
+function getTarget(args) {
+  const args1 = parseInt(args[1], 10)
+  if (Number.isInteger(args1) && args1 < 0) {
+    let coords = '-'
+    for (let i = 0; i < args.length; i++) {
+      if (args[i] === '-,') {
+        coords += ','
+        continue
+      }
+
+      const uint = args[i].substring(1)
+      if (!Number.isInteger(parseInt(uint, 10))) {
+        continue
+      }
+
+      if (args[i - 1] === '--') {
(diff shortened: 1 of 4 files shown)


src/commands/info.ts
```diff
@@ -37,12 +37,41 @@ export const help = () => `
       ${chalk.green('$ dcl info 0x8bed95d830475691c10281f1fea2c0a0fe51304b')}
 `
 
+function getTarget(args) {
```
m-sossich: Why not define a  _--target_ parameter?  I mean instead of doing this custom arg parsing.

Something like: dcl info --target=-18,40

fmiras: For many reasons:
1) I think that we have to design the command regardless of the technologies that we're using, and in this case, I think that is more intuitive to set the target as an "arg" instead of as an "option".
2) Flags in CLI are use `--like this` and not `--like=this` to support that we should change `arg` as dependency and find another library or make a big refactor or just make a ugly parsing similar like this one.

src/commands/info.ts.rej
```diff
@@ -0,0 +1,5 @@
+diff a/src/commands/info.ts b/src/commands/info.ts	(rejected hunks)
+@@ -76,2 +78 @@ export async function main() {
+-    contentServerUrl:
+-      args['--host'] || 'https://content.decentraland.org'
++    contentServerUrl: args['--host'] || 'https://content.decentraland.org'
```
menduz: ?

fmiras:
@menduz good point! https://github.com/decentraland/cli/issues/352

PR #350, opened 2019-01-11
fix: ensure that the process exit after command

### LINKED ISSUE #347: Exit process on deployment success
Currently the process is not exited on deployment success

fmiras:
Fixes #347

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -131,4 +131,4 @@ async function main() {
   }
 }
 
-main()
+main().then(() => process.exit(0))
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -131,4 +131,4 @@ async function main() {
   }
 }
 
-main()
+main().then(() => process.exit(0))
```
menduz:
```suggestion
main()
  .then(() => process.exit(0))
  .catch((err) => {
    // do something with the error
    process.exit(1)
  })
```

Issue #354, opened 2019-01-13
Deployments that overlap previous scenes have various unexpected results


cfravel:
--- Environment
Windows 10
CLI 2.0.1-20190111190610.commit-547cd83
https://client.decentraland.today/?position=58%2C147

--- Scenes used in these tests:
Scene A:  an 8-parcel "donut" with SW/base parcel at 57,147  
   (a nine parcel square with missing central parcel)
   It has a standard static boilerplate in the 57,147 parcel, the other 7 parcels are empty.

Scene B: a 1-parcel "donut hole" at 58,148.  Static boilerplate, but the sphere is blue

Scene C: a 9 parcel square (donut+hole) deployed at 57,147.  '
  Standard static boilerplate in 57,147 but the ball is dark red/brown,
  and there is a bright green sphere in 58,148

Scene D1: a 4-parcel scene with 4 instances of the static boiler plate, SW/base parcel is at 58,149
   but each parcel also has a lowpoly mesh gltf-embedded big number (1, 2, 3, 4)
   this scene is organized with having 4 "entity" tags wrapping each of the parcels' contents.

Scene D2: same as D1 but deployed with its SW/base parcel at 56,147

--- Steps to reproduce
1. Deploy A: success (you can upload a donut)

2. Deploy B: success, (you can place a scene into a donut hole)

3. Deploy D1: warns about overlap.  Yes, replace.
// Is this normal: the content in the sw parcel of A didn't disappear.   
// That is a parcel in an overlapping scene, but that particular parcel isn't part of the overlap.

4. Deploy C (9 parcel scene that replaces both A and B). Warns about overlap.
//  As expected, the dark red and green balls appear, and the donut hole content with the blue sphere disappears.

5. Deploy B (to replace the "hole" of scene C.   I don't recall whether it warned of overlap.
// Unexpected: Scene B (blue ball etc does NOT replace the green ball of Scene C
// This is probably a bug.

6. Deploy D2
// it shows up, replacing the red ball etc of C
// green ball is still present in 58,148
// It seems that generally replacements happen on a parcel by parcel basis, an overlapping scene does not replace all of the old scene.

7. Deploy A (donut), overlapping with D2.   
// Unexpected: no warning of overlap
// Unexpected:  Its content at 57,147 does not replace the content in that parcel from D2
// Unexpected: the blue ball from B shows up again!  (the hole in A allows some old content to reappear)  this is probably a bug.

cfravel:
[...]

PR #366, opened 2019-01-16
feat: add a warning message on overriding scenes

fmiras:
Refer to #354


twaldorf:
Small spelling + copy suggestion for clarity:

> Deploying this scene will override any other overlaped one

to

> Deploying this scene will override and break any overlapping scenes

fmiras:
@twaldorf why do you think that "break" is a better word? The client aren't going to render the previous/overlapping scenes at all, not even partially or with errors.

Issue #374, opened 2019-01-22
dcl start throws errors for ecs scenes in builds after 2.1.0-20190117174453.commit-68f71cd


cfravel:
--- environment
Windows 10, Chrome

---- The following works --------
npm i -g decentraland@2.1.0-20190117174453.commit-68f71cd
dcl init
dcl start

----- If I use 2.2.0, 2.2.1 or 2.2.2 I get the following:
dcl init
dcl start
√ Installing dependencies

> dcl-project@1.0.0 watch C:\Users\cfrav\webvr\DCL\ECS-boilerplate-2019-01-21
> build-ecs --watch

'build-ecs' is not recognized as an internal or external command,
operable program or batch file.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! dcl-project@1.0.0 watch: `build-ecs --watch`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the dcl-project@1.0.0 watch script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm WARN Local package.json exists, but node_modules missing, did you mean to install?

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\cfrav\AppData\Roaming\npm-cache\_logs\2019-01-22T05_45_13_000Z-debug.log
Error: `dcl start` Main scene file bin/game.js is missing, run "dcl help start" for more info.

fmiras:
Please @cfravel check again, I see in this output that you're executing `dcl init` and `dcl start` before the dependencies get installed. It should be:
```bash
dcl init
√ Installing dependencies
dcl start
```
And not:

```bash
dcl init
dcl start
√ Installing dependencies
```

cfravel:
i will retest tonight.
I doubt that I could be doing dcl start before the dependencies are installed, because I don't get the command line cursor back.
I just didn't put it in the report.

menduz:
We have a hint about the issue, we will post an update soon, Thanks

cfravel:
Did a clean install on another laptop today
Got the same error

C:\Users\cfravel>npm i -g decentraland
C:\Users\cfravel\AppData\Roaming\npm\dcl -> C:\Users\cfravel\AppData\Roaming\npm\node_modules\decentraland\bin\dcl
+ decentraland@2.2.2
added 1 package in 1.369s

C:\Users\cfravel>md webvr

C:\Users\cfravel>cd webvr

[...]

Issue #357, opened 2019-01-14
feat: improve error handling


fmiras:
# What? 
As today most of the errors are handled properly but the messages that are being displayed are the same as the errors expose. We should save that errors and show them only in debug mode and present a generic friendly message by default

# Why? 
Improve developer experience

PR #376, opened 2019-01-22
fix: handle errors of npm install

fmiras:
Related #374

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -3,7 +3,8 @@ import * as arg from 'arg'
 import chalk from 'chalk'
 import opn = require('opn')
 
-import { loading, info, warning } from '../utils/logging'
+import * as spinner from '../utils/spinner'
+import { warning } from '../utils/logging'
 import { Analytics } from '../utils/analytics'
 import { ErrorType, fail } from '../utils/errors'
 import { Decentraland } from '../lib/Decentraland'
@@ -60,7 +61,7 @@ export async function main() {
   dcl.on('link:ready', url => {
     Analytics.sceneLink()
     console.log(chalk.bold('You need to sign the content before the deployment:'))
-    const linkerMsg = loading(`Signing app ready at ${url}`)
+    spinner.create(`Signing app ready at ${url}`)
 
     setTimeout(() => {
       try {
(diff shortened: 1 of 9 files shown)


menduz:
Looks good, how do we test it?

src/commands/deploy.ts
```diff
@@ -84,15 +85,15 @@ export async function main() {
   })
 
   dcl.on('upload:starting', () => {
-    const uploadMsg = loading(`Uploading content...`)
+    spinner.create(`Uploading content...`)
 
     dcl.on('upload:failed', (error: any) => {
-      uploadMsg.fail('Fail to upload content')
+      spinner.fail('Fail to upload content')
```
michaltakac: Grammar nazi here, "Failed"

src/index.ts
```diff
@@ -148,7 +148,6 @@ export async function main(version: string) {
       )
     )
     debug(e)
-    await Analytics.reportError(e.name, e.message, e.stack)
```
menduz: Where did this go?

src/utils/moduleHelpers.ts
```diff
@@ -14,14 +17,33 @@ export function setVersion(v: string) {
...
+    throw new Error('Unable to install dependencies: no internet connection')
+  }
+
+  return installDependencies(silent)
+}
+
+export async function installDependencies(silent: boolean = isDebug()): Promise<void> {
+  spinner.create('Installing dependencies')
   return new Promise((resolve, reject) => {
     const child = spawn(npm, ['install'], { shell: true })
```
menduz: probably setting the cwd of the project will help

fmiras: why do you think that? remember that the `process.cwd()` gives the path where the command was executed and not the working dir (since the commands can receive the path as parameter: e.g. `dcl init my-folder/`

src/commands/deploy.ts
fmiras: Great, thanks!

src/index.ts
fmiras: I removed it since it has very inaccurate information (like the error type, messages and stack, and also it's blocking the process a lot. The next thing on the list is re-doing this in https://github.com/decentraland/cli/issues/357

src/utils/moduleHelpers.ts
menduz: that is the reason, you need to pass `my-folder` to the cwd

fmiras: Oh sorry, I know what you mean, thanks.

PR #379, opened 2019-01-25
chore: use consistent current working directory

menduz:
This PR fixes some inconsistencies in the resolution of the current working directory


menduz:
Let's merge this to see if it works now.

PR #394, opened 2019-02-15
feat: preview - only prompt metamask on eth-connect scenes

### LINKED ISSUE #388: Metamask is always required to open a preview scene
Today, you always need to sign with Metamask to open a scene preview. Even if the scene doesn't use the blockchain at all.

This is a big point of friction when sharing content, because most people don't have Metamask (and don't even know what it is).

It would be great to remove that requirement, at least for scenes that don't have eth-connect or Ethereum controller imported.

fmiras:
Closes #388 

Related/dependant: https://github.com/decentraland/client/pull/561

# What?
Pass as script variable the boolean of `node_modules/eth-connect` path exists to the client preview artifact 

# Why?
So users don't have to handle the overhead of web3, having metamask installed or having to logging/signing if they don't need it

### DIFF
--- a/src/lib/Preview.ts
+++ b/src/lib/Preview.ts
@@ -116,9 +116,28 @@ export class Preview extends EventEmitter {
       )
     }
 
-    this.app.get('/', (req, res) => {
+    this.app.get('/', async (req, res) => {
       res.setHeader('Content-Type', 'text/html')
-      res.sendFile(path.resolve(artifactPath, 'artifacts/preview.html'))
+      const ethConnectExists = await fs.pathExists(
+        path.resolve(this.dcl.getWorkingDir(), 'node_modules', 'eth-connect')
+      )
+
+      const htmlPath = path.resolve(artifactPath, 'artifacts/preview.html')
+
+      if (!ethConnectExists) {
+        return res.sendFile(htmlPath)
+      }
+
+      const html = await fs.readFile(htmlPath, {
+        encoding: 'utf8'
+      })
+
+      const response = html.replace(
+        'src="/@/artifacts/preview.js"',
+        `src="/@/artifacts/preview.js" ethConnectExists="${ethConnectExists}"`
+      )
+
+      res.send(response)
     })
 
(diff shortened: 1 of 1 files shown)


src/lib/Preview.ts
```diff
@@ -116,9 +116,28 @@ export class Preview extends EventEmitter {
       )
     }
 
-    this.app.get('/', (req, res) => {
+    this.app.get('/', async (req, res) => {
       res.setHeader('Content-Type', 'text/html')
-      res.sendFile(path.resolve(artifactPath, 'artifacts/preview.html'))
+      const ethConnectExists = await fs.pathExists(
+        path.resolve(this.dcl.getWorkingDir(), 'node_modules', 'eth-connect')
```
menduz: Maybe we should search for `@decentraland/EthereumController` since you may request only the public key to identify the user

src/lib/Preview.ts
```diff
@@ -116,9 +116,28 @@ export class Preview extends EventEmitter {
...
+        return res.sendFile(htmlPath)
+      }
+
+      const html = await fs.readFile(htmlPath, {
+        encoding: 'utf8'
+      })
+
+      const response = html.replace(
+        'src="/@/artifacts/preview.js"',
+        `src="/@/artifacts/preview.js" ethConnectExists="${ethConnectExists}"`
```
menduz: Take a look at the other PR (client). We can maybe redirect to `?NO_ETHEREUM` or something so we don't have to monkeypatch the response

src/lib/Preview.ts
fmiras: The thing is that, if the scene evens need to identify the users, it will use eth-connect and if that so this is not it's path. This is for non-blockchain interaction at all.

src/lib/Preview.ts
fmiras: It may be a good resolution but If they want this scene on now for example it'll be `my-awesome-scene.com?NO_ETHEREUM` (this is not that cool)

PR #397, opened 2019-02-20
fix: dcl start models aren't correctly mapped

fmiras:


src/lib/Preview.ts
```diff
@@ -139,12 +139,8 @@ export class Preview extends EventEmitter {
           res.json(err)
           res.end()
         } else {
-          files
-            .filter($ => !ig.ignores($))
-            .filter($ => fs.statSync($).isFile())
-            .map(relativiseUrl)
-
-          const ret = getDummyMappings(files)
+          const ret = getDummyMappings(files.map(relativiseUrl))
```
eordano: This might be a separate issue, but we should be using this function from the standard library!!!

 https://nodejs.org/api/path.html#path_path_relative_from_to

menduz: It is super specific for our usage and to work only with URLs path.relative works in more obscure ways depending on the OS and doesn't take into account all the domain logic. Neither URL.relative works for our use case

PR #398, opened 2019-02-20
 feat: suggest building on deploy/export fail because of missing script

### LINKED ISSUE #373: Suggest building project when deploy failed
# What? 
Sometimes users try to `dcl deploy` before building since they do it before even running `dcl start` for the first time (which it runs `npm run build` for us). We should suggest running `npm run build` in case the reason of deploy failure is that `game.js` file is missing
 
# Why? 
So users get to know how to solve this issue.

fmiras:
Closes #373

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -6,6 +6,8 @@ import chalk from 'chalk'
 import opn = require('opn')
 
 import * as spinner from '../utils/spinner'
+import buildProject from '../utils/buildProject'
+import isECSProject from '../utils/isECSProject'
 import { warning } from '../utils/logging'
 import { Analytics } from '../utils/analytics'
 import { ErrorType, fail } from '../utils/errors'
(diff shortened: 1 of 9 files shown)


src/commands/deploy.ts
```diff
@@ -58,6 +59,18 @@ export async function main() {
 
   const workingDir = args._[1] ? path.resolve(process.cwd(), args._[1]) : process.cwd()
 
+  const sceneJson = await fs.readJSON(path.resolve(workingDir, 'scene.json'))
+  const mainPath = path.resolve(workingDir, sceneJson.main)
+
+  if (!(await fs.pathExists(mainPath))) {
```
menduz: If it doesn't exist and it requires a doable precondition. Can we just run build for them,?

fmiras: But the thing is that people may want to export without building, because of changes on assets, or JSON data, etc. Also automatically building on each command make the possible building pipeline very slow, think of doing `dcl start` && `dcl export`, we're building the project twice and It get slower. I though of automatically building but then I choose the next.js aproach of just suggesting building because the error could be something else

eordano: Can we use `make` as intended? This was exactly why `make` was built: check the date of the built file, and if the source files have been modified at a later date than the built file, then and only then trigger the build.

menduz: I don't get it, are you saying we should use make from the CLI to leverage the outdated file checks? Make doesn't work in windows

matiasbargas: From what I understand, we should run build only when the `mainPath` doesn't exist instead of notifying the user. If it already exists we should deploy as it is.  `start` should always build and show the latest updated version in the preview.

fmiras: I like @matiasbargas proposal, @menduz @eordano thoughts?

PR #401, opened 2019-03-11
fix: show export command on help

fmiras:


src/index.ts
```diff
@@ -42,6 +42,7 @@ const help = `
       init                Create a new Decentraland Scene project
       start               Start a local development server for a Decentraland Scene
       deploy              Upload scene to Decentraland
+      export              Export scene to static content
```
nearnshaw: I'd give a little more context to what we mean by this. Also "static" could be interpreted as that the scene isn't interactive.
How about:
"Export scene to content that's compatible with 3rd party servers. Ex:Zeit Now, GitHub pages, and Amazon S3."

fmiras: It's too ambiguous, the Dockerfile approach is also compatible with 3rd party servers or the `dcl start --ci` command, the CLI is prepared for both unexperienced and experienced developers, this is only a very short description that have to be very specific on what the command does we could change it by "static files" or "" instead of "static content" but "compatible with 3rd party servers", isn't even what it does. This kind of information should be available, of course, at the docs but I think that the help command should have more straightforward tech/low-level definitions. 

Also about "static scenes" isn't a very cool description, maybe we should start populating other kind of name such as "non-interactive scenes" or something else. 

Another description that could be even more accurate is: "Export scene to serverless content". What do you think of that?

PR #420, opened 2019-05-08
feat: use real CIDs in /mappings

menduz:
To correctly count textures


nearnshaw:
Textures are now counted well. This scene from the builder used to break because of the texture count, now it only counts 2. This makes sense according to Shibu, some asset packs use a different texture from others.

Screen Shot 2019-05-08 at 18 33 37

Materials count is still high, 51 in this case. According to Shibu, all Builder assets share a same material.

nearnshaw:
Screen Shot 2019-05-08 at 18 43 12

Another scene with the same, textures have gone down dramatically to just 2, which is awesome
Material count is still higher than it should, but at least not over the scene limits

PR #431, opened 2019-06-02
feat: expose wallet, set a different working dir

Alonzo-Coeus:
This pull will add an optional option to Decentraland.ts allowing for a wallet instances to be passed in.
It will also allow for the working dir to be change by a function call.

I would like this merged so my repo Alonzo-Coeus/dcl-cli-api can hook into the cli and allow us to run the deploy command for multiple directories without resorting to the shell.


fmiras:
Will close since `dcl` is not working as a library, we're probably releasing a library to do all the stuff that CLI does but programmatically and I'll update this issue

Issue #440, opened 2019-06-10
Provide a /scenes and /parcel_info endpoint for the new content API


eordano:
See:

https://github.com/decentraland/content-service/commit/791d7da5402a4faa706bd0437c2cb8e14f8e90bb

PR #442, opened 2019-06-13
feat: add mock scene endpoints

eordano:
# What?
Add a `/scenes` and `/parcel_info` endpoints according to the latest https://github.com/decentraland/content-service changes

# Why?
So we can unblock https://github.com/decentraland/explorer/pull/224

Closes https://github.com/decentraland/cli/issues/440

### DIFF
--- a/src/lib/Preview.ts
+++ b/src/lib/Preview.ts
@@ -133,13 +133,13 @@ export class Preview extends EventEmitter {
     })
 
     this.app.get('/scenes', async (_, res) => {
-      const mapping = watcher.getMappings()[0]
+      const mapping = watcher.getMappings()
       const { parcel_id, root_cid, publisher } = mapping
       return res.json({ data: [{ parcel_id, root_cid, publisher, scene_cid: '' }] })
     })
 
     this.app.get('/parcel_info', (_, res) => {
-      const mapping = watcher.getMappings()[0]
+      const mapping = watcher.getMappings()
       const { parcel_id, root_cid, publisher } = mapping
       return res.json({
         data: [
--- a/src/lib/Preview.ts
+++ b/src/lib/Preview.ts
@@ -128,8 +128,30 @@ export class Preview extends EventEmitter {
 
     this.app.use(nocache)
 
-    this.app.get('/mappings', (req, res) => {
+    this.app.get('/mappings', (_, res) => {
       res.json(watcher.getMappings())
+    })
+
+    this.app.get('/scenes', async (_, res) => {
+      const mapping = watcher.getMappings()[0]
+      const { parcel_id, root_cid, publisher } = mapping
+      return res.json({ data: [{ parcel_id, root_cid, publisher, scene_cid: '' }] })
+    })
+
+    this.app.get('/parcel_info', (_, res) => {
+      const mapping = watcher.getMappings()[0]
+      const { parcel_id, root_cid, publisher } = mapping
+      return res.json({
+        data: [
+          {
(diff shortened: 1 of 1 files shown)


src/lib/Preview.ts
```diff
@@ -128,10 +128,28 @@ export class Preview extends EventEmitter {
 
     this.app.use(nocache)
 
-    this.app.get('/mappings', (req, res) => {
+    this.app.get('/mappings', (_, res) => {
       res.json(watcher.getMappings())
     })
 
+    this.app.get('/scenes', async (_, res) => {
+      const mapping = watcher.getMappings()[0]
```
menduz: What is `getMappings()[0]`? Is it trying to get the scene.json? Isn't it better to filter by `scene.json`?

src/lib/Preview.ts
```diff
@@ -128,10 +128,28 @@ export class Preview extends EventEmitter {
...
+      const { parcel_id, root_cid, publisher } = mapping
+      return res.json({ data: [{ parcel_id, root_cid, publisher, scene_cid: '' }] })
+    })
+
+    this.app.get('/parcel_info', (_, res) => {
+      const mapping = watcher.getMappings()[0]
```
menduz: Same here

eordano: ditto

src/lib/Preview.ts
eordano: Because this is the preview, `mappings` always returns 1 "mappings". This endpoint takes advantage of that fact and ignores the `x1`, `y1`, `x2`,... parameters that are sent.

This is a first step towards the new `scene.json`, where we still don't have the mappings inside the scene.json.

menduz: my coment about what is the [0] is because getMappings does not return an array

```ts
  public getMappings() {
    const contents: Array<{ file: string; hash: string }> = []

    this.mappings.forEach((hash, file) => contents.push({ file, hash }))

    return {
      contents,
      parcel_id: '0,0',
      publisher: '0x0000000000000000000000000000000000000000',
      root_cid: this.mappings.get('scene.json') || 'Qm0000000000000000000000000000000000000000'
    }
  }
```

src/lib/Preview.ts
```diff
@@ -128,10 +128,28 @@ export class Preview extends EventEmitter {
 
     this.app.use(nocache)
 
-    this.app.get('/mappings', (req, res) => {
+    this.app.get('/mappings', (_, res) => {
```
fmiras: Shouldn't this endpoint be deleted?

eordano:
On it.

src/lib/Preview.ts
menduz: No, to provide support to older version of the SDK

menduz:
@eordano please revert the deletion of /mappings, it will disable completely older versions of the SDK and break projects that wasn't upgraded yet

eordano:
@menduz given that the fix for that would be to upgrade the SDK, wouldn't that be OK?

The only case where that would break is if somebody did `npm i decentraland-ecs@latest` but did not `npm i -g decentraland`, right?

menduz:
We better keep both for a couple of versions until we can seamlessly migrate the users

src/lib/Preview.ts
menduz: Resolved via https://github.com/decentraland/cli/pull/442/commits/61cc38bd58b70bfa93b179e92fa9986211053ba6

PR #479, opened 2019-08-27
feat: show friendly message on deploy error

### LINKED ISSUE #465: Bug deploying: new BigNumber() not a base 16 number
Several people, me included, have had this bug when trying to deploy: 

new BigNumber() not a base 16 number

I don't know of any pattern for when it happens. In my case, it happens when trying to deploy some scenes, not others.

If I do this `export DCL_PRIVATE_KEY=<your wallet priv. Key>` ,  I can deploy those same scenes, so it seems not to be related to the scene's content.

At least the error message should say something clear and actionable in these cases.

fmiras:
Closes #465

### DIFF
--- a/linker-app/src/components/LinkerPage/LinkerPage.tsx
+++ b/linker-app/src/components/LinkerPage/LinkerPage.tsx
@@ -2,7 +2,7 @@ import * as React from 'react'
 import { Address, Blockie, Header, Button } from 'decentraland-ui'
 import Navbar from 'decentraland-dapps/dist/containers/Navbar'
 
-import { baseParcel, rootCID, isRopsten } from '../../config'
+import { baseParcel, rootCID, isRopsten, isDebug } from '../../config'
 import Error from '../Error'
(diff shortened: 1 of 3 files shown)


linker-app/src/components/LinkerPage/LinkerPage.tsx
```diff
@@ -129,7 +129,13 @@ export default class LinkScenePage extends React.PureComponent<LinkerPageProps,
             check the CLI for more info.
           </p>
         )}
-        {error ? <Error>{error}</Error> : null}
+        {error ? (
+          isDebug() ? (
+            <Error>{error}</Error>
+          ) : (
+            <Error>There was an unexpected error.</Error>
```
nicosantangelo: Just a suggestion, maybe we can encourage an action here. I don't know if something like "refresh" will help, but something like that.

fmiras: https://github.com/decentraland/cli/issues/486 🔢

PR #482, opened 2019-09-05
chore: use `webpack` instead of `ncc`

### LINKED ISSUE #368: Fix duplication of files at building
# What? 
When using `npm run cli:build` or `npm run cli:watch` scripts the ncc static evaluator or mod resolving is also duplicating all the `lib` folder from `src` to `dist`. We should either delete it after building or change ncc configuration to stop doing that, same with assets such as `certs` or `css`.

# Why? 
Avoid adding extra files to the build and get a smaller install size npm package.

fmiras:
Fixes #368


eordano:
To provide some context, we use protobuf as the serialization mechanism for
our real time comms. Not sure if it's needed on the client though!

PR #489, opened 2019-09-18
chore: update comms proto

moliva:
Fixes https://github.com/decentraland/kernel/issues/493.

# What? 
Update comms & broker protocols and test server.

# Why? 
- Updates both protocol to match the current ones used by kernel & comms service
- Fixes issue regarding multiplayer scenes that are not been run appropriately as of today

**NOTE:** Might need to check retrocompatibility or advise users to update their SDK version.


eordano:
:( we really shouldn't have generated files in our repo :(

moliva:
We might need to move this to a shared dependency to avoid these issues.

PR #491, opened 2019-10-14
feat: first working version of deploy with WC

fmiras:
# What? 
Use alternative deploy with Wallet Connect compatible wallet instead of linker-dapp


src/commands/deploy.ts
```diff
@@ -186,7 +191,39 @@ export async function main() {
     }
   }
 
+  if (args['--wallet-connect']) {
+    const walletConnector = new WalletConnect({
+      bridge: 'https://bridge.walletconnect.org' // Required
```
nicosantangelo: Why? 😛

src/commands/deploy.ts
```diff
@@ -186,7 +191,39 @@ export async function main() {
...
+
   const address = await dcl.deploy(files)
   Analytics.sceneDeploySuccess({ address })
   return console.log(chalk.green(`\nDeployment complete!`))
 }
+
+async function connect(walletConnector: WalletConnect): Promise<void> {
+  return new Promise((resolve, reject) => {
+    walletConnector.on('connect', (error, payload) => {
+      console.log('connected', error, payload)
```
nicosantangelo: Is this log intended? (sorry I'm not _that_ familiar with the codebase)

src/walletConnect/connector.ts
```diff
@@ -0,0 +1,15 @@
+import { WalletConnect } from '.'
+
+let connector: WalletConnect
+
+export function isWC() {
```
nicosantangelo: Why not using WalletConnect here?
the other methods already do:
```suggestion
export function isWalletConnect() {
```

src/index.ts
```diff
@@ -1,3 +1,4 @@
+WebSocket = require('ws')
```
nicosantangelo: Why's this here? we should maybe leave a comment?

src/lib/Decentraland.ts
```diff
@@ -288,6 +289,17 @@ export class Decentraland extends EventEmitter {
...
     }
 
+    if (isWC()) {
+      const connector = getWalletConnector()
+      const address = connector.accounts[0]
+      const signature = await connector.signMessage([address, messageToSign])
+      return {
+        signature,
+        address,
+        network: { id: connector.networkId, name: 'mainnet' }
```
nicosantangelo: Is mainnet the only option in our particular use-case?

src/walletConnect/index.ts
```diff
@@ -0,0 +1,10 @@
+import Connector from '@walletconnect/core'
+import { IWalletConnectOptions } from '@walletconnect/types'
+
+import * as crypto from './crypto'
+
+export class WalletConnect extends Connector {
+  constructor(opts: IWalletConnectOptions) {
+    super(crypto, opts)
```
nicosantangelo: **Question**: Do you have to implement the `crypto` module each time you want to implement this connector ?
Seems odd that's all (and scary)

src/commands/deploy.ts
fmiras: Why is required? I leaved that comment that is in the quick start documentation but the reason is just because of how WC works. It needs a bridge server that we could host but is easier to use theirs so we don't have to maintain it (exactly as all the things we use with infura) 

Related: https://walletconnect.readthedocs.io/en/latest/technical-specification/#bridge-api-reference

src/commands/deploy.ts
fmiras: No, good catch. I'm going to fix it

src/commands/deploy.ts
nicosantangelo: Ok!

does it make sense to leave this link/mini-explanation there? it's just weird to specify that that option is required in particular

src/walletConnect/index.ts
fmiras: Yes :|

This is the part that should be an npm package to support WC in terminal/node.js

PR #498, opened 2020-03-08
feat: build entity and exit

eordano:


nearnshaw:
What's the entity.json file for?
Why would you only want to build that file?

nchamo:
Closing due to inactivity

PR #502, opened 2020-03-31
fix: Show error when deployment fails

nchamo:
We are now showing the cause of the deployment failure. This is because we are now using the new `dcl-catalyst-client`.


src/commands/deploy.ts
```diff
@@ -13,17 +12,15 @@ import { Decentraland } from '../lib/Decentraland'
...
+import { debug } from '../utils/logging'
 
 export const help = () => `
   Usage: ${chalk.bold('dcl build [options]')}
 
     ${chalk.dim('Options:')}
 
       -h, --help                Displays complete help
-      -t, --target              Specifies the address and port for the target content server. Defaults to https://peer.decentraland.org/content
+      -t, --target              Specifies the address and port for the target content server. Defaults to https://peer.decentraland.org
```
marcosnc, line 23: Be sure to not add the content suffix automatically, remember that we may want to deploy to a local server or a server running outside our catalyst nodes.

src/commands/deploy.ts
```diff
@@ -13,17 +12,16 @@ import { Decentraland } from '../lib/Decentraland'
...
 
 export const help = () => `
   Usage: ${chalk.bold('dcl build [options]')}
 
     ${chalk.dim('Options:')}
 
       -h, --help                Displays complete help
-      -t, --target              Specifies the address and port for the target content server. Defaults to https://peer.decentraland.org/content
+      -t, --target              Specifies the address and port for the target catalyst server. Defaults to peer.decentraland.org
+      -t, --target-content      Specifies the address and port for the target content server. Can't be set together with --target
```
marcosnc, line 24: Can we add an example for this option, please?

PR #518, opened 2020-09-03
fix: Reverting catalyst client library to fix deploy

pablitar:
# What? 

`Error: dcl deploy window is not defined, run “dcl help deploy” for more info.`

According to @nchamo there is a dependency that got added/updated when updating the catalyst client that is causing the problem.

Reverting the catalyst-client should fix it for now.


marcosnc (approved):
Please keep an eye on this to go back to version 2.X as soon as we find a fix for the underlying issue @nchamo.

PR #523, opened 2020-10-07
feat: enable debugging in vscode

menduz:
# What? 
* Prepare the ground for the new debugger functionalities.
* Fixes comms errors after every update

Once the PR is merged and the explorer#1404 also is, the way to access the new debugger is by adding `?WS_SCENE=1`to the preview URL.

Important: This is an opt-in feature. This is NOT the default behavior.

# Why? 
Level up in the maturity of the development cycle 🚀

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1886,73 +1886,6 @@
         }
       }
     },
-    "@ledgerhq/devices": {
-      "version": "4.68.4",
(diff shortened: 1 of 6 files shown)


nchamo (approved):
Great work!

Could we add some docs on how to use it?

cc @nearnshaw

menduz:
Sure do, it certainly depends on the explorer PR to work. I'd like to keep it private until it is stable tho.

PR #524, opened 2020-10-08
fix: "dcl build" should always show the output of the build

menduz:
# What? 
Should always show the output of the build. I had a scene not working and it was because it had errors in the code I was not seeing.

# Why? 
Avoid headaches

### DIFF
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -30,7 +30,7 @@ export async function main(): Promise<number> {
   const workDir = process.cwd()
 
   if (await isTypescriptProject(workDir)) {
-    await buildTypescript(workDir, !process.env.DEBUG, !!args['--watch'])
+    await buildTypescript(workDir, false /* silent=false build should always inform errors */, !!args['--watch'])
   }
 
   return 0


nchamo (approved):
Great work!

src/commands/build.ts
```diff
@@ -30,7 +30,7 @@ export async function main(): Promise<number> {
   const workDir = process.cwd()
 
   if (await isTypescriptProject(workDir)) {
-    await buildTypescript(workDir, !process.env.DEBUG, !!args['--watch'])
+    await buildTypescript(workDir, false /* silent=false build should always inform errors */, !!args['--watch'])
```
sandrade-dcl, line 33: Only to try to follow a standard in the team about write comments in the code: Wouldn't be cleaner to put these kind of comments out of the function signature? What do you think?

sandrade-dcl (changes requested):

src/commands/build.ts
menduz: I tend to put what a boolean value is right next to it. Otherwise unless you are reading code in a typescript capable editor you won't know what a wild "false" is in the code.

sandrade-dcl: Can't we put this comment on top of the call to the function? or even inside the `buildTypescript()` function as header of it?

For example in C# we write documentation for functions in this way:

`/// <summary>`
`///  Description for the function...`
`/// </summary>`
`/// <param name="payload">Description for param...</param>`

But I don't know what would be the correct way for typescript code.

sandrade-dcl: Anyway, I understand you have to move forward with it and we can discuss this kind of things in the future ;)

sandrade-dcl (approved):

src/commands/build.ts
BrianAmadori: I'do it like this to make the comment a bit more readable:
```
await buildTypescript(
    workDir, 
    false, // silent=false build should always inform errors 
    !!args['--watch'])
```

We can also add the comment to the signature using this format already present in SDK code:
```
  /**
   * Description...
   * @param silent - when setting this to false the build should always inform errors.
   */
```

PR #529, opened 2020-12-15
fix: show error when scene project doesn't have tsconfig.json

nchamo:
Previously, we failed silently when a project didn't have a `tsconfig.json`. Now we show the error message


marcosnc (approved):

menduz (approved):

menduz:
Return 1 may be good or simply "throw"

PR #530, opened 2021-01-27
fix: update libs in order to support Metamask changes

nchamo:
Main changes:
* Updated `decentraland-dapps`
* Started using `ethers` instead of `decentraland-eth`
* Started using webpack for building the linker app, instead of the `decentraland-compiler`


pablitar (approved):
Great work!!

Issue #533, opened 2021-02-21
'dcl info' has stopped working, throws an error


cfravel:
This is hindering district management of tenants.

nchamo:
Hey @cfravel !

We just released a new `@next` version, that you can install by doing:
```npm install -g decentraland@next```

Could you confirm that the fix works for you? If it does, then we can release it as a new version

cfravel:
I've examined this for 73,92.  This is land in Aetheria whose UpdateManager, UpdateOperator and LandUpdateOperator have all been set.
I've also examined it for 14,100 in the Conference Center

DCL INFO is now partially fixed:
1. It doesn't throw an error
2. In both the LAND and ESTATE sections it gets Owner right (including the Passthrough contract, e.g. for Aetheria that is 0x494b9f97c63dfc8c6e6a4298a114e82a2e28e895
3. In the LAND section It gets UpdateOperator right. For 73,92 this is correctly x4959f54f7b30224047ac3cae6bf147bbaee8b61c 
4. In the ESTATE section it gets the UpdateOperator right.   For all of the Aetheria estate, this is 0x98b39480da71e6355f704d67d36f443085af1d8f and for all of the Conference Center this is 0xaf293617a79cCC47958aB781e62fc8c886C8536c
5. It now does show information for 151,151, which is one of the parcels of the Aetheria estate that lies ouside of 150,150 and does have the Estate UpdateOperator set and shown properly

Not correct or missing.
6. It doesn't report anything for Operator in either LAND or ESTATE sections.  I don't know what field this is supposed to be, but it is blank in both sections
7. It doesn't display an important ESTATE field: UpdateManager (that should show as an Ethereum contract address for Aetheria, and for 14,100 in the Conference Center the UpdateManager should show as 0x9609d68eAB639349e6E9a2eec76839ec8d109C22
8. The parcels listed used to be all the parcels in the estate.  Now it is some random list of parcels.

6 and 7 are not new issues, i believe, but 7 would be important to fix for estate management.
8. Is a new breakage with this @next version, and should be fixed asap, hopefully before deployment to latest

cfravel:
Yikes, another case that is still broken

9. dcl info still throws an error if you run it on a parcel that isn't in an estate and doesn' have a scene deployed (e.g. -1,144)

nchamo:
Reopening to continue working on the fixes

@cfravel , could you provide an example for 8 (where the parcels looks random), and specify what the list of parcels should be

[...]

PR #534, opened 2021-02-23
fix: use subgraph since API v1 was removed

nchamo:
Two different services used by the command `dcl info` were recently removed:
* The old content server
* The v1 marketplace API

We are updating the CLI to use the new content server, and one of dcl's subgraphs

Fixes https://github.com/decentraland/cli/issues/533


marcosnc (approved):
Great work!

PR #535, opened 2021-02-25
fix: support estates with more than 100 parcels

nchamo:
This is a continuation of https://github.com/decentraland/cli/pull/534, trying to solve remaining issues in https://github.com/decentraland/cli/issues/533

We are no supporting estates with more than 100 parcels, and stop failing when information about an empty parcel is required

### DIFF
--- a/src/lib/API.ts
+++ b/src/lib/API.ts
@@ -79,20 +79,24 @@ export class API implements IEthereumDataProvider {
     return response.estates[0].updateOperator
   }
 
+  // This is a special case, because some estates have +10000 parcels, and TheGraph doesn't support offsets of more than 5000
   async getLandOfEstate(estateId: number): Promise<Coords[]> {
-    const query = `query GetLandOfEstate($estateId: String!) {
-      estates(where:{ id: $estateId }) {
-        parcels {
+    const query = `query GetLandOfEstate($estateId: String!, $lastId: String!) {
+      estates(where: { id: $estateId }) {
+        parcels (where:{ id_gt: $lastId }, first: 1000, orderBy: id) {
           x
           y
+          id
         }
       }
     }`
-    type ResultType = { estates: { parcels: { x: string; y: string }[] }[] }
-    const response = await this.fetcher.queryGraph<ResultType>(dclApiUrl, query, {
-      estateId: `${estateId}`
-    })
-    return response.estates[0].parcels.map(parcel => ({
+    type ResultType = { estates: { parcels: { x: string; y: string; id: string }[] }[] }
+    const response = await this.queryGraphPaginated<
+      ResultType,
+      { x: string; y: string; id: string }
+    >(query, { estateId: `${estateId}` }, result => result.estates[0].parcels)
+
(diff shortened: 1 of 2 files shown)


src/lib/content/ContentService.ts
```diff
@@ -34,8 +34,12 @@ export class ContentService extends EventEmitter {
    * @param y
    */
   async getSceneData(coordinates: Coords): Promise<SceneMetadata> {
-    const entity = await this.fetchEntity(coordinates)
-    return entity.metadata
+    try {
+      const entity = await this.fetchEntity(coordinates)
+      return entity.metadata
+    } catch {
+      return null
```
nchamo, line 41: It used to return `null` before https://github.com/decentraland/cli/pull/534, so we are continuing to do the same

src/lib/API.ts
```diff
@@ -79,20 +79,24 @@ export class API implements IEthereumDataProvider {
...
-    type ResultType = { estates: { parcels: { x: string; y: string }[] }[] }
-    const response = await this.fetcher.queryGraph<ResultType>(dclApiUrl, query, {
-      estateId: `${estateId}`
-    })
-    return response.estates[0].parcels.map(parcel => ({
+    type ResultType = { estates: { parcels: { x: string; y: string; id: string }[] }[] }
+    const response = await this.queryGraphPaginated<
+      ResultType,
+      { x: string; y: string; id: string }
+    >(query, { estateId: `${estateId}` }, result => result.estates[0].parcels)
```
marcosnc, line 97: May `result.estates[0]` be undefined?

marcosnc (approved):
Great work!

src/lib/API.ts
```diff
@@ -149,29 +153,61 @@ export class API implements IEthereumDataProvider {
...
+
+    const finalResult: ArrayType[] = []
+    while (keepGoing) {
+      const result = await this.fetcher.queryGraph<QueryResultType>(dclApiUrl, query, {
+        ...variables,
+        first: pageSize,
+        lastId
+      })
+      const array = extractArray(result)
+      keepGoing = array.length === pageSize
```
agusaldasoro, line 206: have you tested this method when the loop executes after the last page is full, but the next page is empty?

agusaldasoro (approved):
Great Work!

src/lib/API.ts
nchamo: Yes, the next page will be empty and array will be empty as well, so it won't fail

agusaldasoro: Will the query handle okay the `lastId` param as undefined? Does ` $lastId: String!` mean that it's mandatory?

nchamo: But for `lastId` to be undefined, then `array` needs to have size 0, and then `keepGoing` would be `false`. Right?

agusaldasoro: Oh, that's true! (I got confused and thought that the `lastId` would be undefined, but it'd be the last element from the previous page instead).

PR #537, opened 2021-03-10
fix: avoid peer as default

nchamo:
Prior to this change, the CLI would use peer.decentraland.org as default, when deploying a new scene.

The problem is that this caused a really big single point of failure.

Now, the CLI will choose a random catalyst that is approved by the DAO

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -1,6 +1,6 @@
 import * as arg from 'arg'
 import chalk from 'chalk'
-import { ContentClient, DeploymentBuilder } from 'dcl-catalyst-client'
+import { CatalystClient, ContentAPI, ContentClient, DeploymentBuilder } from 'dcl-catalyst-client'
 import { EntityType } from 'dcl-catalyst-commons'
 import { Authenticator } from 'dcl-crypto'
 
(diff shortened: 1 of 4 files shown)


src/commands/deploy.ts
```diff
@@ -117,23 +117,25 @@ export async function main(): Promise<number> {
...
+      spinner.create(`Uploading data to: ${target}`)
     } else if (args['--target-content']) {
-      contentServerAddress = args['--target-content']
+      const targetContent = args['--target-content']
+      catalyst = new ContentClient(targetContent, 'CLI')
+      spinner.create(`Uploading data to: ${targetContent}`)
     } else {
-      contentServerAddress = 'peer.decentraland.org/content'
+      catalyst = await CatalystClient.connectedToCatalystIn('mainnet', 'CLI')
+      spinner.create(`Uploading data to a random catalyst on the network`)
```
marcosnc, line 135: Are we showing the selected random server at some point?

marcosnc (approved):
Great Work!

src/commands/deploy.ts
nchamo: Nope. That's is not exposed by the library yet. I'll create an issue to add that to the library so that we can show it here

agusaldasoro (approved):
Great Work!

webpack.config.js
```diff
@@ -12,7 +12,11 @@ module.exports = {
     new webpack.BannerPlugin({
       banner: '#!/usr/bin/env node',
       raw: true
-    })
+    }),
+    new webpack.ProvidePlugin({
+      window: 'global/window',
+   }),
```
menduz, line 18: Why is that even necessary for the cli bundle?

nchamo: Basically this

https://github.com/webpack/webpack/issues/7112#issuecomment-703187668

PR #562, opened 2021-06-17
fix: avoid errors with 0 value

nchamo:
Fixes https://github.com/decentraland/cli/issues/533

**Context:**
`dcl info` can show information about a specific parcel. It prints on the console some information regarding the scene's metadata.

The problem was that a specific scene had a `0` set as a spawn point. So the code would not enter the `if (item)` part, and the accumulator in the reduce would be undefined (since there was no else). In the next iteration, it would fail

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "decentraland",
-  "version": "3.5.0",
+  "version": "3.5.1",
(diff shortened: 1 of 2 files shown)


pablitar (approved):
LGTM!

PR #563, opened 2021-06-22
chore: make source maps go away for production builds

menduz:
# What? 

...

# Why? 

...


nearnshaw:
This fixes things in the case that people manually run `dcl build` before deploying

They often will not. Running `dcl start` is considered equivalent to running `dcl build`.

Ideally, `dcl deploy` should also run `build-ecs --production` in the background
(this would have other extra benefits BTW)

PR #568, opened 2021-07-29
chore: make example scene easier to use

nchamo:
Fixes https://github.com/decentraland/sdk/issues/17

I even added a default thumbnail 😄


nearnshaw (commented):
How about we add one more field, so it's clearer where this needs to go in case of being used

```
  "featureToggles": {},
```

nearnshaw (approved):

PR #571, opened 2021-08-05
feat: import setupExport from decentraland-ecs and call it if exists

leamendoza97:
# What? 
Find a setupExport.js file in the decentraland-ecs package, if it exists, it's called it instead of standard export.

# Why? 
In accordance with the new package structure, this allows setting up a custom package export from the ECS. Also, it's maintaining compatibility with the current structure.

# How to test?
1. Build and link the CLI `npm run build && npm link`
2. In a scene project, run `dcl export` and check if it works properly
3. Remove the export generated and create a file in `node_modules/decentraland-ecs/src/setupExport.js` with `module.exports = function (exportDir, sceneTitle, mappings){
    console.log("Hello world");
}`
4. Run again `dcl export` and note only project files were copied without any package

### DIFF
--- a/src/commands/export.ts
+++ b/src/commands/export.ts
@@ -1,3 +1,4 @@
+import { MappingsFile } from './../lib/content/types'
 import * as path from 'path'
 import * as fs from 'fs-extra'
 import * as arg from 'arg'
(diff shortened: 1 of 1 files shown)


src/commands/export.ts
```diff
@@ -69,46 +71,64 @@ export async function main(): Promise<number> {
...
-      path.resolve(artifactPath, 'images/progress-logo.png'),
-      path.resolve(exportDir, 'images/progress-logo.png')
-    ),
-    fs.copy(
-      path.resolve(artifactPath, 'images/teleport.gif'),
-      path.resolve(exportDir, 'images/teleport.gif')
-    )
-  ])
+  if (externalExport) {
+    externalExport(exportDir, sceneJson.display.title, mappings)
```
menduz, line 89: To this function I would pass an object instead of ordered arguments, here is the rationale: https://menduz.com/posts/2019.05.07#at-root-level-always-return-objects-never-arrays

It makes the signatures more extensible.

Also, please add `await` to enable async operations

What I am imagining is to passing something like this:

```ts
await externalExport({
  workingDir,
  exportDir,
  // sceneJson, // <-- this can be inferred from $workingDir/scene.json if necessary.
  mappings
})

File: src/commands/export.ts [hunk 2]
```diff
@@ -69,46 +71,64 @@ export async function main(): Promise<number> {
...
+      ),
+      fs.copy(
+        path.resolve(artifactPath, 'images/progress-logo.png'),
+        path.resolve(exportDir, 'images/progress-logo.png')
+      ),
+      fs.copy(
+        path.resolve(artifactPath, 'images/teleport.gif'),
+        path.resolve(exportDir, 'images/teleport.gif')
+      )
+    ])
```
menduz, line 130: This could be extracted to a function

menduz (changes requested):

src/commands/export.ts
leamendoza97: I already changed this but sceneJson. We only need to replace the Html's title. The sceneJson is loaded in the CLI with dedicated helpers for it (not big deal, json reader, types and something else), and that is the reason why I included it.

src/commands/export.ts
leamendoza97: Done

menduz (approved):

Issue #567, opened 2021-07-22
Create `dcl install <package>`


menduz:
Due to the bug described in: https://github.com/npm/cli/issues/3174

1. Create a process that goes through all the installed dependencies and looks for `decentralandLibrary` in `<lib>/package.json` and adds them to `bundleDependencies`.
2. Create `dcl install <package>` to execute `npm install` and then execute the process from (1)

PR #573, opened 2021-08-09
feat: make dcl install

leamendoza97:
# What? 
Add the command **install** to install decentraland libraries and to check all decentraland libraries installed which need to be in bundleDependencies field of _package.json_

- [x] Main functionality, look for all dependencies and select decentralandLibraries
- [x] Install a new package
- [x] Tests

# Why? 
See #567


src/main.ts
```diff
@@ -38,6 +38,7 @@ const help = `
       init                Create a new Decentraland Scene project
       build               Build scene
       start               Start a local development server for a Decentraland Scene
+      install             Install a package and/or sync bundleDependencies
```
leamendoza97, line 41: Maybe two lines with 
- install
- install package
?

menduz (approved):
Awesome, only one question. Are the `␊` in the new lines required?

leamendoza97:
> Awesome, only one question. Are the `␊` in the new lines required?

That file is auto-generated by the `npm run test:dry` to update the snapshots which match in the CI running so, I think it's okay.

menduz:
🚢 it

PR #576, opened 2021-08-13
feat: add --web3 arg to dcl start 

leamendoza97:
# What? 
Add extra arguments `--web3` to `dcl start` which adds `&ENABLE_WEB3` query param to preview URL and the auto-execute of browser

# Why? 
https://github.com/decentraland/sdk/issues/26


menduz (approved):

src/commands/start.ts
```diff
@@ -26,6 +26,7 @@ export const help = () => `
       -b, --no-browser          Do not open a new browser window
       -w, --no-watch            Do not open watch for filesystem changes
       -c, --ci                  Run the parcel previewer on a remote unix server
+      --web3                    Add &ENABLE_3 to preview URL and first running
```
nearnshaw, line 29: New creators won't know what this parameter is for. I would make the explanation about why you would want to use this parameter.

Something like:
Connects preview to browser wallet to use the associated avatar and account

nearnshaw (approved):

PR #585, opened 2021-09-20
feat: add github actions

gonpombo8:
# What? 

Add github actions workflows.

...

# Why? 
We want to remove circle ci, and use github actions instead
https://github.com/decentraland/sdk/issues/61
...


menduz (commented):
Adding to the description-less issue https://github.com/decentraland/sdk/issues/61:
- Can you set up the workflow to run in PRs as well in master? `oddish-action` does not publish on pull requests, only in tags, master and main branches
- It would be really cool if we could remove entirely appveyor and circle altogether, probably a matrix setup using both windows and mac would be good for this PR

.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: npm run test:win
+      - name: test
+        if: matrix.platform != 'windows-latest'
+        run: npm run test:ci
+
+  publish:
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
+      - uses: actions/checkout@master
```
menduz, line 37: I think you may need to build again here. Because jobs do not automatically persist the builds in Actions

```suggestion
      - uses: actions/checkout@master
      - name: Use Node.js 14.x
        uses: actions/setup-node@v1
        with:
          node-version: 14.x
      - name: install
        run: npm install
      - name: build
        run: npm run build
```

menduz (approved):
Looking good, please confirm if my comment applies or not

menduz:
:shipit:

.github/workflows/ci.yml
gonpombo8: 👌🏼

PR #587, opened 2021-09-25
refactor bundledDependencies to read package.json instead of npm ls

gonpombo8:
# What? 
Replace npm ls with readFile(package.json)

...

# Why? 
npm ls exit with code 1 if there were unmet peerDependencies.
...


leanmendoza (approved):

src/utils/installedDependencies.ts
```diff
@@ -1,65 +1,61 @@
...
-): Promise<{ dependencies: ModuleItem[] }> {
-  let dependenciesList: ModuleItem[] = []
-  const dependencyTree = (await getDependencyTree(workingDir)) as { dependencies: Object }
+export async function getDecentralandDependencies(
+  dependencies: Record<string, string>,
+  workDir: string
+): Promise<string[]> {
+  const dependenciesName = []
+  for (let dependencieName of Object.keys(dependencies)) {
+    const modulePath = getPath(workDir, dependencieName)
```
menduz, line 50: Just checking if `require.resolve(dependencyName + '/package.json', { paths: [workDir] })` is a better candidate to resolve paths here

src/utils/installedDependencies.ts
```diff
@@ -1,65 +1,61 @@
...
-  workingDir: string
-): Promise<{ dependencies: ModuleItem[] }> {
-  let dependenciesList: ModuleItem[] = []
-  const dependencyTree = (await getDependencyTree(workingDir)) as { dependencies: Object }
+export async function getDecentralandDependencies(
+  dependencies: Record<string, string>,
+  workDir: string
+): Promise<string[]> {
+  const dependenciesName = []
+  for (let dependencieName of Object.keys(dependencies)) {
```
menduz, line 49:
```suggestion
  for (let dependencyName of Object.keys(dependencies)) {
```

src/utils/installedDependencies.ts
gonpombo8: webpack require.resolve returns the module id instead of the path 😢 
(https://webpack.js.org/api/module-methods/#requireresolve)

src/utils/installedDependencies.ts
gonpombo8: 😂

src/utils/installedDependencies.ts
menduz: Omg, makes you question why a node tool has webpack in the first place

PR #598, opened 2021-11-08
add warning when the scene metadata is the default one

gonpombo8:
`Closes https://github.com/decentraland/sdk/issues/88`
# What? 

https://github.com/decentraland/sdk/issues/88
...

# Why? 
To avoid deploying scenes without thumbnail, title or description
...


menduz (changes requested):
I would take the chance to add this validation: https://github.com/decentraland/common-schemas/blob/main/src/platform/scene/scene.ts#L123

menduz (approved):

PR #603, opened 2021-11-25
feat: types and px feature flag

leanmendoza:
# What? 
Use the `@dcl/schemas` and add the --px temporal flag for `dcl init`

# Why? 

...


src/lib/Project.ts
```diff
@@ -37,6 +38,10 @@ export class Project {
     this.workingDir = workingDir
   }
 
+  async getInfo(): Promise<ProjectInfo> {
```
gonpombo8, line 41: do we need this getInfo fn ? Are we using it ?

gonpombo8 (commented):
I think that this PR has some linter errors.
Already create a PR to run the linter at the ci, but meanwhile can we fixed them here ?

src/lib/Project.ts
leanmendoza: Not really, but I'd want to add it to use it in the `js-sdk-toolchain` and avoid redefining the same function.

leanmendoza:
> I think that this PR has some linter errors. Already create a PR to run the linter at the ci, but meanwhile can we fixed them here ?

We can wait the other PR

src/lib/Project.ts
gonpombo8: 👍🏼 nice

gonpombo8 (approved):

PR #622, opened 2022-01-10
feat: add workspaces to preview many projects at the same time

leanmendoza:
# What? 
- Add multiple projects support. 
- Add a CI to test the CLI before @next version
- Add `export` deprecation message.
- Prioritize localhost or 127.0.0.1 urls for Windows users. 

# Approach

In this approach the workspace consists in a workspace file description and a package.json with the `decentraland-ecs` package to serve the preview mode. 
Each project needs its own `node_modules` directory because can be different version each other.

Now:
- [x] Flexible code to prepare the path of future work
- [x] Mutate class to functions in new things
- [x] Each project with its own node_module to compile the project, but not to run preview (serve renderer and kernel)
- [x] Workspace folder has its own node_module to server renderer, kernel and the preview index html file.
- [x] Workspace json file has the vs code schema
- [x] Prioritize the backwards compatibility, but workspace feature only works from a workspace (or global in the future) ECS at least 6.9.0
- [x] Hot reloading working

Future work:
- [ ] Workspace dependency-less, using a global updated `decentraland-ecs` package, this allows that the workspace file be a single file and can be anywhere (like vs code workspace file)
- [ ] With previous ticked, an extension to install these global dependencies, like node, npm, CLI and decentraland-ecs. As default a project can use the global `decentraland-ecs` but the priority would be to load the local folder ECS

# Other repository dependencies
https://github.com/decentraland/js-sdk-toolchain/pull/89
https://github.com/decentraland/kernel/pull/170


src/lib/Workspace.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export class Workspace {
+  private workingDir: string
+  private projects: Project[] = []
+
+  constructor(workingDir: string) {
+    this.workingDir = workingDir
+
+    const projectFolders = this.getProjectFolders()
+    if (projectFolders.length > 0) {
```
gonpombo8, line 14:
```suggestion
    if (projectFolders.length) {
```

src/commands/start.ts
```diff
@@ -70,59 +68,79 @@ export async function main() {
...
+
+    const [sdkOutdated, ECSInstalled] = await Promise.all([
+      getOutdatedApi(project.getProjectWorkingDir()),
+      isECSInstalled(project.getProjectWorkingDir())
+    ])
+
+    if (!ECSInstalled) {
+      spinner.info('SDK not found. Installing dependencies...')
+    } else if (sdkOutdated) {
+      spinner.warn(
```
gonpombo8, line 98: should we add some log info about the scene/project that has the error ?
To avoid searching which one is the outdated

src/lib/Workspace.ts
```diff
@@ -0,0 +1,78 @@
+import path from 'path'
+import fs from 'fs-extra'
+import { readJSONSync } from '../utils/filesystem'
+import { Project } from './Project'
+
+export class Workspace {
```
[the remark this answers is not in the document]
leanmendoza: Sure!

leanmendoza:
PR description just updated!

gonpombo8 (approved):

PR #646, opened 2022-03-03
Removing unused dependencies

juanigallo:
Fixes https://github.com/decentraland/sdk/issues/186

# What? 

Removing the following dependencies as they are not being used:

- request
- husky
- global

There are some more improvements that can be done in terms of reducing package size as removing isomorphic-fetch (https://github.com/nodejs/node/pull/41749) but this was a low hanging fruit.

Build is still working:

image

All tests are still passing

image

# Why? 

`npm install` time is pretty slow


gonpombo8 (changes requested):
It's seems that u are using a version of `node <= 16`, since u change the entire package-lock.json and it removes all the peerDependencies and changed the lockFileVersion to 1 also.
For example if u try to run eslint with this package-lock.json, will throw an error `sh: 1: eslint: not found`
Reset the package-lock file, install node 16.x, and run npm install again to fix this errors.

Thanks for the PR 🙌🏼

juanigallo:
> 

Sure, will do. If that is the reason I think we need to update the docs (https://docs.decentraland.org/development-guide/SDK-101/). Up to yesterday the required node version in there was 8 and I updated to 14 because the CLI is using optional chaining.

gonpombo8:
> > 
> 
> Sure, will do. If that is the reason I think we need to update the docs (https://docs.decentraland.org/development-guide/SDK-101/). Up to yesterday the required node version in there was 8 and I updated to 14 because the CLI is using optional chaining.

You can use node version 14 or less if u want for the cli. Every command will work, but for contributing/developing we use node 16 with package-lock version 2. There are two diff things, one for content creators, and the other for contributors.

menduz (approved):

juanigallo:
> > > 
> > 
> > 
> > Sure, will do. If that is the reason I think we need to update the docs (https://docs.decentraland.org/development-guide/SDK-101/). Up to yesterday the required node version in there was 8 and I updated to 14 because the CLI is using optional chaining.
> 
> You can use node version 14 or less if u want for the cli. Every command will work, but for contributing/developing we use node 16 with package-lock version 2. There are two diff things, one for content creators, and the other for contributors.

@gonpombo8  Awesome, so maybe adding

```
  "engines": {
    "node": ">=16"
  },
```

in package.json? Or adding in into the README in [Contributing ](https://github.com/decentraland/cli#contributing). Im trying to think which is the best way for outside contributors to know this before submitting a PR

gonpombo8:
> > > > 
> > > 
> > > 
> > > Sure, will do. If that is the reason I think we need to update the docs (https://docs.decentraland.org/development-guide/SDK-101/). Up to yesterday the required node version in there was 8 and I updated to 14 because the CLI is using optional chaining.
> > 
> > 
> > You can use node version 14 or less if u want for the cli. Every command will work, but for contributing/developing we use node 16 with package-lock version 2. There are two diff things, one for content creators, and the other for contributors.
> 
> @gonpombo8 Awesome, so maybe adding
> 
> ```
>   "engines": {
>     "node": ">=16"
>   },
> ```
> 
> in package.json? Or adding it into the README in the [Contributing ](https://github.com/decentraland/cli#contributing) section. Im trying to think which is the best way for outside contributors to know this before submitting a PR

sure ! Will do 
Thanksss

gonpombo8:
Closed via https://github.com/decentraland/cli/pull/674
Thanks for the heads up of the unnecessary deps :)

PR #660, opened 2022-03-28
chore: check node version and update linker dapp package

leanmendoza:
# What? 

...

# Why? 

...

### DIFF
--- a/test/failNodeVersion.js
+++ b/test/failNodeVersion.js
@@ -1,0 +1,11 @@
+const minNodeVersionDev = 16;
+const nodeVersion = process.versions.node;
+if (nodeVersion) {
+  const majorVersion = parseInt(nodeVersion.split(".")[0]);
+  if (majorVersion < minNodeVersionDev) {
+    console.error(
+      `Error: Decentraland CLI is developed over node version ${minNodeVersionDev} or greater, current is ${majorVersion}.\n`
+    );
+    process.exit(1);
+  }
+}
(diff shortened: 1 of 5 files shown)


test/failNodeVersion.js
```diff
@@ -0,0 +1,11 @@
+const minNodeVersionDev = 16;
```
gonpombo8, line 1: Why not still using the `packageJson.engines` and read the values from there ?
Also we should check the npm version.

I think the best solution is to create a fn that receives the expected "engines" and fail/succed comparing it with the npm/node process version.

This way we can reuse this fn for the pre-install and also when we run dcl start

gonpombo8 (changes requested):

test/failNodeVersion.js
gonpombo8: Also, could we move this script to a `.ts` file and run it with ts-node ? 😄

leanmendoza: Do you mean a libraryless and plain javascript function? (which can be executed with pure node)
Because in the pre-install stage, you don't have Typescript yet. And the `dcl [command]` is a post-built CLI functionality.

leanmendoza: Moving the script to a `.ts` makes typescript necessary to install:
- Dev case: they didn't install anything yet
- User case: I think is fine, because the `.ts` is already compiled in `dist/file.js`

leanmendoza: I'm thinking aloud

PR #662, opened 2022-03-30
feat: check node&npm version and deprecate export

leanmendoza:
# What? 
- Add checks for node and npm version for users
- ~~Fail~~ **Warn** install when contributors run `npm i` in the CLI repo (removing .npmrc) 
- DEPRECATE `dcl export`

# Why? 
- About deprecation: this will be deprecated in ECS 7 (this means won't there be a setupExport.js file)


package.json
```diff
@@ -106,5 +106,9 @@
     "node": ">=16.0.0",
     "npm": ">=8.0.0",
     "yarn": "please use npm"
+  },
+  "userEngines": {
+    "minMajorNode": 14,
```
gonpombo8, line 111: what if we use the same schema as "engines" and then use semver.lt to compare the versions?

src/utils/nodeAndNpmVersion.ts
```diff
@@ -0,0 +1,41 @@
+import { spawn } from 'child_process'
+
+// This file tries not having dependencies
+
+export const npm = /^win/.test(process.platform) ? 'npm.cmd' : 'npm'
+
+export async function getNodeMajorVersion(): Promise<number | boolean> {
+  const nodeVersion = process.versions.node
```
gonpombo8, line 8: we could return the process.versions.node and then use semver to compare the literal versions.

test/unit/utils/gettingPackageJsonAndVersioning.ts
```diff
@@ -0,0 +1,30 @@
+import test from 'ava'
+import { getCLIPackageJson } from '../../../src/utils/moduleHelpers'
+import {
+  getNpmMajorVersion,
+  getNodeMajorVersion
+} from '../../../src/utils/nodeAndNpmVersion'
+
+test('Unit - getCLIPackageJson() - should have userEngines', async (t) => {
```
gonpombo8, line 8: 🙌🏼  🎉

Issue #664, opened 2022-04-03
Because dcl init now has GIT dependencies, GIT installation should be described in the Installation Guide


cfravel:
At least one of the new CLI scene type options (lazy load) in `dcl init` has a dependency on GIT.  
I believe it would helpful therefore to mention that GIT installation be described, at least as an option, in the SDK Installation Guide.

leanmendoza:
https://github.com/decentraland/cli/pull/673

PR #673, opened 2022-04-26
feat: remove the git dependency in `dcl init`

leanmendoza:
# What? 

...

# Why? 
https://github.com/decentraland/cli/issues/664


src/utils/download.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  const extractedPath = path.resolve(dest, directoryCreated[0])
+  const filesToMove = await readdir(extractedPath)
+
+  for (const filePath of filesToMove) {
+    await move(
+      path.resolve(extractedPath, filePath),
+      path.resolve(dest, filePath)
+    )
+  }
```
menduz, line 42: what is the purpose of this code? does it work recursively with folders?

leanmendoza: Move the content of the extracted folder to its parent. It could be implemented with `fs-extra.moveSync(extractedPath, dest)` but it doesn't work on Linux

src/utils/download.ts
```diff
@@ -1,13 +1,12 @@
 import extract from 'extract-zip'
 import { rmdirSync } from 'fs'
 import { move, readdir, remove, writeFile } from 'fs-extra'
-import * as makeFetchHappen from 'make-fetch-happen'
```
menduz: make-fetch-happen uses local caching 😢

src/utils/download.ts
```diff
@@ -0,0 +1,45 @@
...
+export const downloadRepoZip = async function (url: string, dest: string) {
+  const zipFilePath = path.resolve(dest, 'temp-zip-project.zip')
+  await downloadFile(url, zipFilePath)
+
+  const oldFiles = await readdir(dest)
+
+  try {
+    await extract(zipFilePath, { dir: dest })
+  } catch (err) {
+    console.error(err)
```
gonpombo8, line 21: return an error so if this happens we can debug it.

src/utils/download.ts
```diff
@@ -0,0 +1,45 @@
...
+    throw err
+  }
+
+  const newFiles = await readdir(dest)
+
+  const directoryCreated = newFiles.filter((value) => !oldFiles.includes(value))
+
+  if (directoryCreated.length !== 1) {
+    throw new Error('asd')
```
gonpombo8, line 30: same here:)

gonpombo8 (approved):

PR #674, opened 2022-04-26
update/remove dependencies

gonpombo8:
# What? 

...

# Why? 

...


test/helpers/sandbox.ts
```diff
@@ -12,13 +11,13 @@ export default function sandbox(fn: CallbackFn) {
...
     const done = () => {
-      rimraf(dir, resolve)
+      fs.remove(dir, resolve)
     }
 
     try {
       fn(dir, done)
     } catch (e) {
-      rimraf(dir, () => reject(e))
+      fs.remove(dir, () => reject(e))
```
menduz, line 20: fs.remove works for recursive directories???

gonpombo8: i just try it and yes, it works :)

menduz: What ensures it will finish? Maybe adding an await?

test/unit/lib/Project.test.ts
```diff
@@ -142,12 +142,14 @@ test('Unit - Project.validateNewProject() - should pass if the working directory
   t.pass()
 })
 
-test('Unit - Project.validateNewProject() - should fail if the working directory contains a scene.json file', async (t) => {
+test.only('Unit - Project.validateNewProject() - should fail if the working directory contains a scene.json file', async (t) => {
```
leanmendoza, line 145: maybe remove the `.only` ?

test/helpers/sandbox.ts
leanmendoza: if it fails or not, it's calling reject in the callback.

leanmendoza (approved):

test/unit/lib/Project.test.ts
gonpombo8: 🫠

PR #692, opened 2022-05-12
feat: use wearable.json instead of asset.json

leanmendoza:
# What? 

...

# Why? 

...

### DIFF
--- a/src/project/projectInfo.ts
+++ b/src/project/projectInfo.ts
@@ -1,22 +1,32 @@
 import path from 'path'
-import fs from 'fs-extra'
+import fs, { readJsonSync } from 'fs-extra'
 import { sdk } from '@dcl/schemas'
+import { ASSET_JSON_FILE, WEARABLE_JSON_FILE } from '../utils/project'
 
 export type ProjectInfo = {
   sceneId: string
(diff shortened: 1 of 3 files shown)


src/project/projectInfo.ts
```diff
@@ -1,22 +1,32 @@
...
 
 export type ProjectInfo = {
   sceneId: string
   sceneType: sdk.ProjectType
 }
 
-export function getProjectInfo(workDir: string): ProjectInfo {
-  const assetJsonPath = path.resolve(workDir, 'asset.json')
+export function getProjectInfo(workDir: string): ProjectInfo | null {
+  const assetJsonPath = path.resolve(workDir, ASSET_JSON_FILE)
```
gonpombo8, line 12: since the old portable experience had the asset.json file, should we check first for portable-experience and wearbale.json files instead of smart-items ?

gonpombo8 (approved):

src/project/projectInfo.ts
leanmendoza: asset.json is used to recognize smart-items and it'll be no longer for smart-wearables
for wearables, we'll have `wearable.json` and optional `builder.json`

PR #703, opened 2022-05-19
refactor linker-dapp

gonpombo8:
# What? 
https://github.com/decentraland/linker-dapp/pull/9
https://github.com/decentraland/opscli/pull/156

...

# Why? 

...

### DIFF
--- a/src/lib/LinkerAPI.ts
+++ b/src/lib/LinkerAPI.ts
@@ -8,6 +8,7 @@ import bodyParser from 'body-parser'
 import portfinder from 'portfinder'
 import { ChainId } from '@dcl/schemas'
 import urlParse from 'url'
+import querystring from 'querystring'
 
 import { getPointers } from '../utils/catalystPointers'
 import { Project } from './Project'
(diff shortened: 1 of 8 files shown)


src/lib/LinkerAPI.ts
```diff
@@ -4,9 +4,10 @@ import { EventEmitter } from 'events'
 import urlParse from 'url'
 import fs from 'fs-extra'
 import express from 'express'
+import cors from 'cors'
 import portfinder from 'portfinder'
-import querystring from 'querystring'
 import { ChainId } from '@dcl/schemas'
+import { getPointers } from '@dcl/opscli/dist/commands/pointer-consistency'
```
menduz, line 10: Opscli is not meant to be used as a library, copy and pasting code may be better? wdyt

src/lib/LinkerAPI.ts
```diff
@@ -99,21 +100,42 @@ export class LinkerAPI extends EventEmitter {
...
+
+    this.app.get('/api/catalyst-consistency', async (req, res) => {
+      const { x, y } = await this.project.getParcelCoordinates()
+      const pointer = `${x},${y}`
+      const chainId = this.project.getDeployInfo()?.linkerResponse?.chainId || 1
+      const network =
+        chainId === ChainId.ETHEREUM_MAINNET ? 'mainnet' : 'ropsten'
+      const value = await getPointers(pointer, network, { log: false })
+
+      res.send(value)
```
menduz, line 138: catch (err) { res.status(500).end() }

src/lib/LinkerAPI.ts
gonpombo8: 👌🏼

PR #772, opened 2022-07-21
update testnet to goerli

gonpombo8:
Needs https://github.com/decentraland/linker-dapp/pull/11

# What? 
Upate testnet to goerli
...

# Why? 
Ropsten being deprecated.
...


src/commands/deploy.ts
```diff
@@ -204,7 +204,7 @@ export async function main(): Promise<void> {
```
gonpombo8, line 204: TODO: Update package.json with the latest linker-dapp version after https://github.com/decentraland/linker-dapp/pull/11 being merged

PR #788, opened 2022-08-04
feat: support both ecs6 and ecs7 project

leanmendoza:
# What? 

...

# Why? 

...


src/utils/analytics.ts
```diff
@@ -121,10 +117,6 @@ async function track(
   }
 
   return new Promise<void>(async (resolve) => {
-    const dclApiVersion = await getInstalledVersion(
```
gonpombo8, line 124: should we add something like ecs version here ?

gonpombo8 (approved):

src/utils/analytics.ts
gonpombo8: or add some analytics to track the new sdk ?

PR #801, opened 2022-08-16
feat: add authorization server

HPrivakos:


menduz (commented):
How is this new argument different than '--target-content'?

src/commands/deploy.ts
```diff
@@ -184,7 +200,7 @@ export async function main(): Promise<void> {
   )
 
   // Uploading data
-  let catalyst: ContentAPI
+  let catalyst: ContentAPI | CatalystClient
```
menduz, line 203: Would you use two different variables with different types instead? To prevent typing oddities

HPrivakos: I reverted that, it was to prevent type issues I had while developing it

menduz:
Does it work if we want to specify both the target server and the authentication service? I'm thinking that a `--disable-validations` would simplify things a lot as an alternative implementaiton

HPrivakos:
> Does it work if we want to specify both the target server and the authentication service? I'm thinking that a `--disable-validations` would simplify things a lot as an alternative implementaiton

No, only one can be specified at a time, and we need to check permissions against the server to avoid users without permissions trying to upload scenes against the middleware server

menduz:
I'd suggest to add the `--skip-validations` instead, it seems to be a simpler approach. The content-servers should only expose two endpoints to upload a scene: `POST /entities` and `POST /available-content`.

It would be something like `dcl deploy -t https://my-content-server --skip-validations`

That would also enable people to upload content to *any* content-server compatible API without performing validations.

HPrivakos:
Moved to skip validations and also updated linker-dapp: [https://github.com/decentraland/linker-dapp/pull/12](https://github.com/decentraland/linker-dapp/pull/12)

menduz (approved):

PR #853, opened 2022-10-06
fix: skip validations on custom content servers

menduz:
# What? 
fix: skip validations on custom content servers

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -58,7 +58,7 @@ export async function main(): Promise<void> {
     '-t': '--target',
     '--target-content': String,
     '-tc': '--target-content',
-    '--skip-validations': String,
+    '--skip-validations': Boolean,
     '--skip-version-checks': Boolean,
     '--skip-build': Boolean,
     '--https': Boolean,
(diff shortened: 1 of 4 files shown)


leanmendoza (approved):

src/commands/deploy.ts
```diff
@@ -85,7 +85,11 @@ export async function main(): Promise<void> {
     workingDir: workDir,
     forceDeploy: args['--force-upload'],
     yes: args['--yes'],
-    skipValidations: !!args['--skip-validations']
+    // validations are skipped for custom content servers
+    skipValidations:
+      !!args['--skip-validations'] ||
+      !!args['--target'] ||
```
gonpombo8, line 91: what if u target to `peer.decentraland.zone` ?
Linker-dapp will show that you have permissions, but will fail on the deploy right ? 🤔

menduz: I'd only keep it for the default case, the only valid validation is against the catalyst server itself, and it can have all validations turned off via env var.

I'd only keep the validation for .org catalysts, this is causing troubles in other configurations

PR #869, opened 2022-10-20
feat: add features needed for editor

cazala:
This PR adds a few features needed for the editor. Each feat has been done in its own commit:

- Added `--skip-install` flag to `dcl init` and `dcl start` commands, this is necessary to avoid using the global `npm`. The editor will handle the dependencies on its own. Commit: 5508477eeaa0030c1254ab595ba8521daeb4fb39.
- Added `--no-browser` flag to `dcl deploy`, this is necessary so we can open the linker webapp from within a VSCode webview. Commit: 2657d12d303d79ba8e62ff6b1c1025d4a3c481f7.
- Added `--port` param to `dcl deploy` to allow configuring the port for the linker webapp. The editor handles the ports for all its web servers so this is necessary to keep that management consistent across webviews. Commit 3634d4ed51a577267b3f8d24426757b473068bab.

The rest of the commits are chores related to bumping the package version and configuring `prettier`.


leanmendoza (approved):
LGTM! Run `npm run test:dry` to regenerate the test snapshots.

src/commands/deploy.ts
```diff
@@ -77,6 +83,11 @@ export async function main(): Promise<void> {
   const workDir = process.cwd()
   const skipVersionCheck = args['--skip-version-checks']
   const skipBuild = args['--skip-build']
+  const noBrowser = args['--no-browser']
+  const port = args['--port']
+  const parsedPort = typeof port === 'string' ? parseInt(port, 10) : void 0
```
gonpombo8, line 88: what this sorcery `void 0` ? 🪄 
its the same as using undefined or what ? Just to know 😄

gonpombo8 (approved):

src/commands/deploy.ts
gonpombo8, line 88: why are we parsing the port twice ?
```ts
const linkerPort = parseInt(args['--port']!, 10)
```

cazala: yes, undefined was making the linter complain :/ it said that the variable undefined did not exist..? i didnt dig much tho i just changed it for void 0 lol, it’s like the undefined from 1996 xd

cazala: maybe it’s a misconfig on eslint on my end, im afk now but i ll check later

cazala: `parseInt` returns a `number`, but it could be `NaN` (ie `parseInt("hola") // NaN`), so the secod line makes sure that the `number` is a valid integer (it could also be replaced for `!isNaN(parsedInt)`)

PR #899, opened 2022-11-17
feat: upgrade linker dapp

cazala:
Upgrade `@dcl/linkder-dapp` to get support for WalletConnect, WalletLink and Fortmatic.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "decentraland",
-  "version": "3.14.0",
+  "version": "3.13.0",
   "description": "Decentraland CLI developer tool.",
   "bin": {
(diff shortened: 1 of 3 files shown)


gonpombo8 (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "decentraland",
-  "version": "3.14.0",
+  "version": "3.13.0",
```
gonpombo8, line 3: is this intentional ?

cazala: yep, I dunno why the minor was `14`, the current minor is `12`. I set it to `13` cos that would be the next minor:
Screen Shot 2022-11-17 at 15 51 16

PR #929, opened 2023-02-09
feat: world-acl command

marianogoldman:


pedrotambo (approved):

src/commands/deploy.ts
```diff
@@ -252,7 +252,8 @@ export async function main(): Promise<void> {
     debug('\n' + error.stack)
     failWithSpinner('Could not upload content', error)
   }
-  return
+
+  process.exit(0)
```
menduz, line 256: 0 means success, we should use the `process.exitCode` by default

```suggestion
  process.exit()
```

marianogoldman: Removed the 0. Yes, it is a successful exit.

src/lib/WorldsContentServer.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.options = args
+    this.options.config = this.options.config || getConfig()
+    this.targetContent = args.targetContent
+    if (process.env.DCL_PRIVATE_KEY) {
+      this.createWallet(process.env.DCL_PRIVATE_KEY)
+    }
+  }
+
+  async link(payload: string): Promise<LinkerResponse> {
+    return new Promise<LinkerResponse>(async (resolve, reject) => {
```
hugoArregui, line 40: this is a bit confusing to me, do you need a promise, can't you just return `linker.link(this.options.linkerPort!, !!this.options.isHttps)`?

marianogoldman: Well... the resolution of the promise is on a `link:success` event received from the `WorldsContentServerLinkerAPI`. Not sure how to re-write it without a promise...

hugoArregui (approved):

PR #936, opened 2023-03-02
remove latest condition for sdk7

gonpombo8:
For the SDK7 we dont use anymore the CLI. So if you want to use an outdated version like 7.0.5 this will always fail.


gonpombo8:
> we also have some checks for this on https://github.com/decentraland/cli/blob/main/src/lib/Project.ts#L543
> 
> do we need to update those too?

we are ok with those cases

PR #941, opened 2023-03-20
feat: add parcel count when sending analytics

marianogoldman:


src/commands/deploy.ts
```diff
@@ -252,8 +254,6 @@ export async function main(): Promise<void> {
     debug('\n' + error.stack)
     failWithSpinner('Could not upload content', error)
   }
-
-  process.exit()
```
menduz, line 256: This is needed because node was leaking an open resource/file/handle preventing the CLI from closing

pentreathm (approved):

aleortega (approved):

PR #942, opened 2023-04-19
Adds skip file size check

### LINKED ISSUE #785: If a deployment is bigger than 100MB, then ask for a double confirmation and
In `dcl deploy`:

If the entire size of the deployment is bigger than 100MB:
1. warn the user about it
2. explain that it is likely to fail, suggest splitting it into several scenes with link at documentation
3. offer an ordered-by-size list of files to be deployed to make them aware of what they are trying to deploy
4. ask for confirmation to continue regardless

aviaisenberg:
# What?

This PR adds --skip-file-size-check to dcl deploy which ignores the 50 MB file size check when deploying to a worlds server using --target-content

# Why? 

The deployer should be able to relax this constraint when deploying to a private worlds server. Scenes with file sizes larger than 50 MB function locally when developing, and should be able to be deployed.

closes #785

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -35,6 +35,7 @@ export const help = () => `
       --skip-version-checks     Skip the ECS and CLI version checks, avoid the warning message and launch anyway
       --skip-build              Skip build before deploy
       --skip-validations        Skip permissions verifications on the client side when deploying content
+      --skip-file-size-check    Skip check for maximum file size when deploying to a target content server. Must be used with --target-content
 
     ${chalk.dim('Example:')}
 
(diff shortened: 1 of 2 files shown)


menduz:
Regardless of the limits of each hosting, downloading and processing assets bigger than 50MB is still a problem for any of the explorers right now. Which kinds of files would be sensible to allow above 50mb?

aviaisenberg:
> Regardless of the limits of each hosting, downloading and processing assets bigger than 50MB is still a problem for any of the explorers right now. Which kinds of files would be sensible to allow above 50mb?

Mainly glb files. With these changes I was able to deploy to a worlds server and the explorer loaded the files when connect, using browser based explorer.

menduz (approved):

menduz:
closes #785

PR #944, opened 2023-05-05
Add more fields to report on scene deploy success event

marianogoldman:


src/commands/deploy.ts
```diff
@@ -244,12 +244,15 @@ export async function main(): Promise<void> {
...
     Analytics.sceneDeploySuccess({
       projectHash: dcl.getProjectHash(),
       ecs: await dcl.workspace.getSingleProject()!.getEcsPackageVersion(),
       parcelCount: parcelCount,
-      coords: baseCoords
+      coords: baseCoords,
+      isWorld: !!sceneJson.worldConfiguration,
+      sceneId: entityId,
+      targetContentServer: catalyst.getContentUrl(),
+      worldName: sceneJson.worldConfiguration?.name
```
menduz, line 255: we have two more ways to deploy worlds. we should (must) measure this server-side and not from the CLI and builder and sdk7-commands

marianogoldman: We are already measuring deployments server-side (perhaps not so many details), but it leaves out all non-foundation servers.
What are the other 2? I also covered this one: https://github.com/decentraland/js-sdk-toolchain/pull/570.
May be I can also use the 3rd one.

menduz: that would be the builder repo

hugoArregui (approved):

PR #999, opened 2024-06-11
feat: update land manager subgraph link

meelrossi:
# What? 
This PR replaces the land manager subgraph link with a new one that targets our own subgraph worker. This worker will handle requests for all subgraphs redirecting to TheGraph studio or satsuma depending on availability

# Why? 
TheGraph hosted service reaches EOL


gonpombo8 (approved):

AlejandroAlvarezMelucciDCL (approved):
LGTM!

